PR #8, opened 2026-02-14
add docker support

mike1237:


ancsemi:
Thanks for putting this together, @mike1237 — merged! I'll be building on top of this with a `.dockerignore`, env var support in the compose file, and docs. Appreciate the contribution 🤙

PR #150, opened 2026-03-13
Expanded permissions system

Jaymus3:
My server operates with multiple users who will functionally administrate, so we desired a little more permission management. I broke it down into 3 permission groups:

1. Manage roles
-  Allows users to edit existing roles, and grant roles to other users that they have channel access to. Admin is still monolithic for granting server-wide permissions
2. Manage server
- Allows users to edit branding, whitelist, persistent invite, cleanup, upload limit, and tunneling.
3. Delete channel
- Allows users to delete channels.

There's a lot of eggs in the manage server basket, so we could target breaking that down into subsections as well.

### DIFF
--- a/public/js/app.js
+++ b/public/js/app.js
@@ -213,8 +213,10 @@ class HavenApp {
     if (this.user.isAdmin || this._hasPerm('create_channel')) {
       document.getElementById('admin-controls').style.display = 'block';
     }
-    if (this.user.isAdmin) {
+    if (this.user.isAdmin || this._hasPerm('manage_roles') || this._hasPerm('manage_server')) {
       document.getElementById('admin-mod-panel').style.display = 'block';
+    }
+    if (this.user.isAdmin) {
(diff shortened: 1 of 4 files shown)


ancsemi:
No, I actually don't have a great email contact... I was using help.havenapp@gmail.com but I guess google decided my activity was suspicious and shut it down 😂 (it may be back up now, but I haven't checked in a few days - sorry to anyone who's tried to contact me there!)

I'm definitely open to collaboration, and I think you're right - refactoring would help a lot with maintenance. I've already done this with app.js to an extent, which I think was over 15,000 lines by the time I got to it lol

Problem is, I've been iterating every single day since launch, making it difficult for others to help build things out. It'll probably be much easier once I get this to a more stable place, but for now I think the easiest way to collab would be to push small PRs, and I'll try to be way more proactive with implementing them early. 

Appreciate your interest and help, and feel free to drop any more suggestions or PRs as they come to you!

Issue #144, opened 2026-03-12
[Feature Idea] Forum Channel Type


ArtyDaSmarty:
**The Idea**
Forum Channels in Haven I could see acting similarly with Discord, with the tags, titles, etc., BUT with the bonus of being able to be a subchannel. With it, I think we could also make forums instead not be split like Subchannels, where you have to scroll through on the sidebar, and instead when you click on the Forum Channel, it shows a little customizable screen where you'd usually see a chatroom that is separated via, "Forums You Are In," and, "Forums You Are Not In". You click on the one you want, and then it brings you to the post like any other chatroom, just with the extra button to see the original post (maybe it automatically pins?). This would benefit us Admins/Mods as well who do not want to see EVERY channel/subchannel known to man in our sidebar. Add in a permission on who can make posts, and it would be golden!

The customizable part I think should primarily just be for how the interface is shown. Discord sort of forces people to have to scroll through each individual one, but I could see something similar to the Discord Discovery menu (image below) being a cool alternative way to have the layout!

<img width="1312" height="844" alt="Image" src="https://github.com/user-attachments/assets/c09a44bd-0240-4fd7-abc7-84d9363b6a7a" />

The idea overall primarily came from how I noticed how I prefer to organize things makes our server look very cluttered, so thought forums could be a perfect alternative, especially with all the things you can do with them, like tags, tag searches, not having them all listed in the sidebar, etc.!

Amnibro:
Forum channels would be the second non-text channel type after voice. Feasible but it really wants to be a phased rollout.

**Schema-side it slots cleanly into the existing channel model:**

\`\`\`sql
ALTER TABLE channels ADD COLUMN type TEXT DEFAULT 'text';     -- 'text' | 'voice' | 'forum' | ...
ALTER TABLE channels ADD COLUMN forum_default_tags TEXT;       -- JSON: ["bug","feature","help"]

CREATE TABLE forum_posts (
  id INTEGER PRIMARY KEY AUTOINCREMENT,
  channel_id INTEGER NOT NULL REFERENCES channels(id) ON DELETE CASCADE,
  title TEXT NOT NULL,
  tags TEXT DEFAULT '',                       -- comma-joined, validated against forum_default_tags
  author_id INTEGER NOT NULL REFERENCES users(id) ON DELETE SET NULL,
  pinned INTEGER DEFAULT 0,
  locked INTEGER DEFAULT 0,
[...]

PR #207, opened 2026-03-27
Forum/Proxy Update

ArtyDaSmarty:
I decided to take a crack at it (with help of some AI, as some of the code was foreign to me, so please feel free to judge the code) and I think I made a pretty neat Forum method. It needs a lot of testing from someone more capable for things like themes, quality of life, etc., but I am hoping this lays some form of ground work! It's worked pretty well on my private server so far!

**UPDATE:** I included a baseline for my proxy idea from way back when!


ancsemi:
Wow, this is A LOT. Have you tested it?

ancsemi:
Same treatment as #208, so this one gets an answer instead of sitting indefinitely.

Closing it. Forum channels are still wanted and #144 tracks that.

Two things stop this specific PR. It carries a forum channel type and a proxy idea together, and your own opening line says the proxy came along by accident. Those are separate features with separate risk, and they need separate reviews. And at 1688 lines across 13 files from March, main has moved far enough that it now conflicts in several places.

The forum half on its own, against current main, is a PR I'd read properly. Channel type, its rendering, the settings it needs, nothing else. That's reviewable in an evening rather than being the kind of thing that gets postponed every time it comes up, which is how it ended up sitting this long.

Both of your PRs closing on the same day isn't a verdict on you. It's that big mixed PRs are the ones that rot, and I'd rather tell you that than keep them open as decoration.

Issue #181, opened 2026-03-18
Request: s3 for media storage


jreoka:
This would be a nice option in the config to use an s3 bucket. (Unless you already can but I couldn't find anything on it)

ancsemi:
S3 support would require the AWS SDK as a dependency and a fair amount of restructuring — every stored file path would need to become an S3 URL, uploads/downloads/deletions would all need to go through the SDK, and there'd need to be a new config section for bucket, region, and credentials. It's doable but it's a meaningful chunk of added complexity for a feature that mostly benefits larger or more advanced deployments.

For now, if you need S3-backed storage, mounting S3 as a local filesystem (s3fs or rclone mount) and pointing Haven's uploads folder at it would actually work without any code changes on our end. Worth considering as a workaround in the meantime. I'm leaving this open in case it gains traction, but it's not something I'm planning to tackle in the near term.

ArtyDaSmarty:
@ancsemi An issue is that, for us using things like Zeabur or similar services, it's not possible or easy to get that (the mount method for S3) up and running (or extremely difficult) via the mounting method, at least from my experience.

I swapped over to Zeabur since a lot less is on my hands to maintain and while they offer a lot of storage with their plans, I'd like to be able to have control over my user's data (or at least a more credible company in terms of security), not Zeabur/the server it is on. Services like MEGA, with their encryption, known security measures, etc., allows that.

TLDR: The mount method is a pain in the ass on some services, including Zeabur from what I can tell. And it'd be nice to put data in a more secure company then leaving it with a company you gotta trust that may be great at one thing, but not known/tested for security.

*I am inexperienced, so I may just not be seeing the tools in my face, but that is just what I think with what I tried/know.*

Amnibro:
Doable as a pluggable storage backend without a big refactor. The trick is that almost all the upload code already funnels through three multer instances (\`upload\`, \`fileUpload\`, plus the avatar variant) and one static handler; abstracting those is mostly mechanical.

**Sketch:**

1. **\`src/storage/index.js\`** with a tiny interface:

\`\`\`js
module.exports = {
  put: async (filename, stream, mimetype) => { /* returns public URL */ },
  delete: async (filename) => {},
  serve: async (filename, res) => { /* streams to response */ },
  exists: async (filename) => bool,
  rename: async (from, to) => {},
};
\`\`\`

[...]

PR #208, opened 2026-03-27
S3 Upload Update

ArtyDaSmarty:
I saw why now that complete S3 conversion would be wayyy too much to just code for the sake of it, so instead, I made it so at least images can be stored, if the Admin permits, onto a S3 instance. All functions SEEM fine from testing, but if you find any hiccups, please let me know!


ancsemi:
This is even more 😂

It'll take me a while to go through it. I'll keep the PR open til I get around to it, but I'd like to focus on bug fixes and smaller QoL before adding such a big mod! 

Stay tuned!

Jaymus3:
Can you try to add meaningful commit messages, or code comments? This is an enormous diff to review and has some misc. edits with no clear justification, like updating the node version to 22. Even if you don't have a full understanding of what the commit is accomplishing, asking the agent you're working with to write a description would help a lot.

Jaymus3:
Instead of scrapping the PR, you could amend your commits with descriptions and force push to your branch. That should update the PR as well. In addition to that, a single commit that just adds comments in certain areas would make this so much easier to parse if you're willing to go the extra mile.

ancsemi:
Coming back to this one properly rather than leaving it sitting for another five months.

Closing it, and it isn't a rejection of S3 support. The idea is good and #181 still tracks it.

The problem is this specific PR. It's 4469 lines across 17 files touching the upload path, which is one of the parts of Haven that gets the most scrutiny, and by your own account most of it came out of an agent rather than out of you. That's an honest thing to have said and I'd rather you said it, but it means nobody can answer questions about it: not you, not me, and not @Jaymus3, who raised the same thing back in March about unexplained changes riding along, like the Node version bump. Main has also moved a long way since, so it conflicts heavily now regardless.

I won't merge code into the upload path that nobody can explain. That would be true of a PR I wrote too.

If you still want S3 in Haven, the version that gets merged looks like this: storage backend selection alone, no forum, no proxy, no version bumps, no drive-by edits. A few hundred lines that you can walk through and defend. Do that against current main and it gets a real review.

Thanks for the work, and for being upfront about where it came from.

PR #5312, opened 2026-04-29
fix(#5307): route channel/role/user/account delete through themed confirm

### LINKED ISSUE #5307: Delete confirmations button text
Is this button text intentional on delete confirmations?

Image
-- 
Image

----
Also, deleting channels opens up this web info box, rather than the popup that the other two deletions use. 
Maybe there's a reason for this, but it seems like inconsistent behavior to me.

Image

Amnibro:
Closes #5307

## Summary

#5307 flagged that channel-delete uses the browser's native `confirm()` "web info box" while role / DM / message deletes use the themed modal — and even where the themed modal is in use, the OK button defaults to "Confirm" which reads oddly on a destructive prompt. This PR walks the remaining user-data delete actions onto the same `_showConfirmModal` everything else uses, and gives the modal a smarter default button label.

## What changed

**`public/js/modules/app-ui.js`**
- **Channel delete (ctx menu).** Was calling `confirm()` twice in a row (`confirm.delete_channel` then `confirm.delete_channel_sure`) — that's the "web info box" the issue calls out. Replaced with a single `_showConfirmModal` whose title carries the warning and whose body carries the "absolutely sure" finality copy. Handler now `async`.
- **Admin moderation `delete-user`.** Was calling `confirm()` mid-handler. Now routes through `_showConfirmModal`. Username is destructured up-front so the outer modal cleanup at the bottom of the handler still runs after the inner modal resolves.
- **Self-delete account.** Same migration.

**`public/js/modules/app-admin.js`**
- **Role delete from main role detail view** and **role delete from the create-role / channel-roles modal.** Both replaced with `_showConfirmModal`.

**`public/js/modules/app-utilities.js`**
- **`_showConfirmModal` smart default.** When `danger: true` and no `confirmLabel` is supplied, the OK button now defaults to `t('messages.delete')` (\"Delete\") instead of `t('modals.common.confirm')` (\"Confirm\"). Existing call sites all pass an explicit label, so this is a no-op for them and only improves any future caller that forgets.

No locale string changes — every existing copy key is reused.

## Out of scope

These call sites still use native `confirm()` and can move in a small follow-up — they're not user-data destructive in the same way:

- `app-channels.js:1350` — webhook delete
- `app-media.js:1423` — bot delete
- `app-ui.js:3189` — backup-file delete

Happy to do that here if you'd prefer one PR.

## Test plan

- [x] `node --check` passes on all three edited modules
- [ ] Right-click a channel → Delete → themed modal appears with warning + "absolutely sure" body and a red **Delete** button → confirm → channel deleted
- [ ] Right-click a channel → Delete → cancel → no socket emit
- [ ] Admin moderation modal → choose "delete-user" → themed confirm appears → both confirm and cancel paths leave the outer admin-action modal closed
- [ ] Settings → Delete Account → themed confirm appears with red **Delete** button
- [ ] Admin → Roles → select role → Delete → themed modal with role name in title
- [ ] Channel-roles modal → role row → Delete → same

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,13 @@ Format follows [Keep a Changelog](https://keepachangelog.com/). Haven uses [Sema
 
 ---
 
+## [Unreleased]
+
+### Fixed
+- **#5307: Delete confirmations were inconsistent — channel delete used the browser's native double-`confirm()` "web info box" while role / DM / message deletes used the themed modal.** All user-data delete actions (channel, role, user, account) now route through the themed `_showConfirmModal`. The two chained channel-delete prompts are merged into one modal that surfaces both the warning copy and the "absolutely sure?" finality copy at once. The themed modal also picks a smarter default button label — when `danger: true` and the caller doesn't pass a `confirmLabel`, the OK button now reads "Delete" instead of "Confirm".
+
+---
+
 ## [3.10.11] — 2026-04-28
 
 ### Fixed
--- a/public/js/modules/app-admin.js
+++ b/public/js/modules/app-admin.js
@@ -3114,8 +3114,13 @@ _renderRoleDetail() {
     });
   });
 
-  document.getElementById('delete-role-btn').addEventListener('click', () => {
-    if (!confirm(t('settings.admin.roles_delete_confirm', { name: role.name }))) return;
+  document.getElementById('delete-role-btn').addEventListener('click', async () => {
+    const ok = await this._showConfirmModal(
+      t('settings.admin.roles_delete_confirm', { name: role.name }),
+      '',
+      { danger: true }
+    );
+    if (!ok) return;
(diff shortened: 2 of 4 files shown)


ancsemi:
Implemented and merged in 3.10.12. All user-data delete actions (channel, role, user, account) now go through the themed modal. The double-chained native channel-delete prompts are combined into one modal with both the warning and finality text. Also tweaked the modal so the OK button defaults to "Delete" instead of "Confirm" when the action is dangerous and no custom label is provided. Thanks for the PR!

ancsemi:
Quick follow-up -- same situation as some other recent PRs. When I said "implemented and merged" I was doing it manually and closing the PR, not using GitHub's merge. Your profile didn't get a contribution for this even though the code is in. That's on me for not knowing the right process, and I'm sorry. Going forward I'll actually merge PRs through GitHub so contributors get the credit they're owed.

PR #5315, opened 2026-04-29
fix(#5248): search DMs locally over the decrypted message cache

### LINKED ISSUE #5248: Search doesn't work in DMs
I think the search in DMs is searching non-decrypted messages rather than decrypted message content

Image

Amnibro:
Closes #5248

## Summary

Searching inside a DM short-circuits with \"search is not available in DMs because messages are end-to-end encrypted\" — which makes the search box look broken to the user even though the messages they want to find are sitting decrypted right in front of them. The server can't `LIKE`-match ciphertext rows (and shouldn't try), but the client already has the decrypted DM messages cached in `_lastRenderedMessages` for rendering. So the search just needs to run there.

## What changed

**\`public/js/modules/app-ui.js\`**
- The 400 ms search-input debounce now branches on the active channel's \`is_dm\` flag. DM channels go to a new \`_searchDmCacheLocally(q)\` helper; everything else still emits \`search-messages\` to the server.

**\`public/js/modules/app-utilities.js\`** — new \`_searchDmCacheLocally(query)\`:
- Walks \`_lastRenderedMessages\`, lowercases content + query, returns up to 50 matches in newest-first order.
- Renders into the existing \`search-results-panel\` markup so result rows look identical to the server-side path: author color, formatted time, \`_highlightSearch\` highlight, click-to-jump via \`_jumpToMessage\`.
- Adds a small \`DM (local)\` chip in the header with a tooltip explaining that scope is \"messages already loaded — scroll up to load more history if you don't see what you're looking for.\"

**\`src/socketHandlers/messages.js\`** — left alone on purpose. The \`is_dm\` short-circuit that emits \`{ results: [], isDM: true }\` stays as defense in depth: a stale client that still calls \`search-messages\` for a DM gets the same friendly empty result it always did.

## What didn't change

- No locale strings — the existing \`search-filter-tag\` class is reused for the \`DM (local)\` chip.
- No new socket events; no server-side changes at all.

## Known scope limit (deliberate, not a regression)

Local search only sees what's already loaded in this view (typically the most recent 50–100 messages — same as what's been rendered). If the user wants to find an old message, they need to scroll up to pull more history first. The chip's tooltip says so. A future PR can add a \"search older history\" affordance that pulls + decrypts more pages on demand.

## Test plan

- [x] \`node --check\` passes on both edited modules
- [ ] Open a DM with messages → type a substring of a known recent message → result row appears with the right author, time, highlight, and click-to-jump
- [ ] Type a substring that isn't in the loaded view → empty results panel, \"DM (local)\" chip still shown
- [ ] Open a non-DM channel → search still goes to the server (unchanged path)
- [ ] In a DM, send a new message containing a unique token, search the token → it appears (because incoming messages are appended to the cache)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,13 @@ Format follows [Keep a Changelog](https://keepachangelog.com/). Haven uses [Sema
 
 ---
 
+## [Unreleased]
+
+### Fixed
+- **#5248: Searching inside a DM previously short-circuited with "search is not available in DMs because messages are end-to-end encrypted" — which made the feature look broken even though the messages were sitting decrypted in front of the user.** The server still can't run `LIKE` against ciphertext rows, but the client already has the decrypted DM messages cached in `_lastRenderedMessages` (the same cache used to render them). Search input now detects when the active channel is a DM and routes the query to a new `_searchDmCacheLocally` helper that scans the cached plaintext and renders results in the existing `search-results-panel` UI with a small `DM (local)` tag in the header so the user understands the scope is "messages already loaded — scroll up to load more". Click-to-jump still works via the existing `_jumpToMessage` path. Server-side DM short-circuit is left in place as defense in depth.
+
+---
+
 ## [3.10.11] — 2026-04-28
 
 ### Fixed
--- a/public/js/modules/app-ui.js
+++ b/public/js/modules/app-ui.js
@@ -1306,7 +1306,14 @@ _setupUI() {
     const q = e.target.value.trim();
     if (q.length >= 2 && this.currentChannel) {
       searchTimeout = setTimeout(() => {
-        this.socket.emit('search-messages', { code: this.currentChannel, query: q });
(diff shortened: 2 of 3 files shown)


ancsemi:
Thanks for the PR! These changes (the DM branch in the search handler and `_searchDmCacheLocally`) were already pulled into main, so I'm going to close this rather than create a double-apply — the conflict on CHANGELOG.md was the only thing blocking a clean merge, and main has moved on a few versions since you opened this. The fix is in and credited in the changelog. Really appreciate the contribution!

ancsemi:
Wanted to follow up and be honest about something -- when I said the changes were "already pulled into main" I was applying them manually and then closing the PR. You ended up with a closed badge instead of a merged one, which means no contribution credit on your profile. I didn't understand at the time that doing it that way was essentially the same as rejecting your PR from GitHub's perspective, and your work genuinely ended up in the codebase. Sorry about that, it wasn't intentional disrespect -- I'm still learning how to handle this properly.

Issue #5255, opened 2026-04-14
Feature Request:  Push to talk toggle


KentuckyFriedBlyat:
Simple idea:  Add a switch on the push to talk feature that toggles how it operates from either mic on/off or hold-to-transmit.

Useful for when you have kids and a bitch wife who wont stop screaming at them.  

Also the keybind recorder doesn't register "just" alt or "just" ctrl, or "just" shift.  (useful when gaming as it lets you xmit without having to move a hand away from WASD or off the mouse.

(also doesn't register extra mouse buttons)

ancsemi:
Fair points. The toggle vs hold-to-talk distinction is a good idea, and I can see how that'd be useful during gaming.

The limitation on standalone modifier keys (just Alt, just Ctrl, etc.) and extra mouse buttons is because Electron's shortcut API flat-out doesn't support them. To make those work I'd need to swap in a low-level input hook library, which is a bigger lift but something I'm open to tackling. No promises on timeline but it's noted.

PR #5319, opened 2026-04-29
feat(#5255): PTT recorder — lone modifiers, mouse buttons, toggle/hold

Amnibro:
Refs #5255

## Summary

Three asks from the issue, all in the keybind recorder used by Settings → Desktop Shortcuts (Haven Desktop only):

1. **Lone modifiers** (just Alt / Ctrl / Shift) — useful while gaming so PTT doesn't pull a hand off WASD.
2. **Extra mouse buttons** (Mouse4 / Mouse5 thumb buttons).
3. **PTT toggle vs hold mode** — replaces the previously-hardcoded \"(toggle)\" label that didn't match what the desktop app actually does.

This PR is the **client side**: capture + UI + persistence to the existing shortcuts config. The actual OS-level registration that has to honor lone modifiers and mouse buttons (and read \`pttMode\` to switch behavior) lives in \`ancsemi/Haven-Desktop\` and is a follow-up PR — Electron's stock \`globalShortcut\` doesn't accept either, so the desktop app needs to switch to a more capable hook (e.g. `uiohook-napi`).

## What changed

**\`public/js/modules/app-platform.js\`** — \`_setupDesktopShortcuts\`:

- **Bare-modifier capture.** The previous handler hard-rejected lone modifiers with `if (['Control','Meta','Alt','Shift'].includes(e.key)) return`. Replaced with a deferred-commit pattern: pressing a modifier alone arms `pendingLoneMod`, and releasing it without ever pressing anything else commits as a bare-modifier accelerator on keyup. Pressing a non-modifier in the meantime falls through to the original combo path and clears the pending state. \`Control\` / \`Meta\` map to Electron's \`CommandOrControl\` string; \`Alt\` and \`Shift\` pass through.
- **Mouse buttons.** New \`mousedown\` listener active during recording. Buttons 0/1/2 (left / middle / right) are intentionally left alone so the user can still click around the page. Buttons 3+ commit immediately as \`Mouse<N>\` (so Mouse4, Mouse5, etc.).
- **Failure UX.** When `setConfig` rejects the binding (older desktop build that doesn't understand `Alt` alone or `Mouse4`), the toast explicitly says the desktop app may need an update — instead of a generic \"in use\" message that misled users.
- **PTT mode select.** New `<select id=\"ptt-mode-select\">` is wired to a new `pttMode` key on the shortcuts config. Default `'hold'`. Persisted via the same `shortcuts.setConfig` IPC.

**\`public/app.html\`**

- The PTT row's hardcoded \"(toggle)\" label is replaced with a `<select>` (hold / toggle).
- Settings hint under the table now mentions that lone-modifier and extra-mouse-button bindings need a recent Haven Desktop build to register.

## What didn't change

- No server-side changes — this is all renderer / IPC.
- No locale strings beyond two new keys (`ptt_hold`, `ptt_toggle_mode`, `binding_hint`); the inline English defaults render until the locale files are updated.
- No CSS additions — reuses `.shortcut-key`, `.shortcut-row`, `.settings-select`, `.muted-text`.

## Test plan

- [x] `node --check` passes on `app-platform.js`
- [ ] In Haven Desktop: Settings → Shortcuts → PTT → Change → press Alt alone, release → \"Alt\" appears in the row (and `setConfig({ ptt: 'Alt' })` is called)
- [ ] Same flow with Ctrl+Shift+P → captures combo as before
- [ ] Same flow with thumb mouse button → \"Mouse4\" / \"Mouse5\" appears
- [ ] Left-click during recording → no commit; the click reaches the page as normal
- [ ] PTT mode select: change to \"toggle\" → `setConfig({ pttMode: 'toggle' })` fires; reload settings → select shows \"toggle\"
- [ ] On an older Haven Desktop build that rejects `Alt` alone → toast \"…desktop app version doesn't support this binding type yet\"
- [ ] In a regular browser (not Electron): the section is already display:none — no regressions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,13 @@ Format follows [Keep a Changelog](https://keepachangelog.com/). Haven uses [Sema
 
 ---
 
+## [Unreleased]
+
+### Added
+- **#5255: PTT recorder accepts lone modifiers, extra mouse buttons, and a toggle/hold mode.** Three asks from the issue, all in the keybind recorder under Settings → Shortcuts (Haven Desktop only). Lone modifiers (just Alt / Ctrl / Shift) now commit on keyup if no other key was pressed in the meantime — useful while gaming so PTT doesn't pull a hand off WASD. A new `mousedown` listener captures buttons 3+ (Mouse4 / Mouse5 thumb buttons); left/middle/right pass through unchanged. The PTT row now has a hold/toggle select that replaces the previously-hardcoded "(toggle)" label, saved via `havenDesktop.shortcuts.setConfig({ pttMode })` (default `hold`). The actual OS-level registration of bare modifiers / mouse buttons and the runtime honoring of `pttMode` are a separate PR in `ancsemi/Haven-Desktop`; older desktop builds get a clear "may need an update" toast when the new binding is rejected by Electron's `globalShortcut`.
+
+---
+
 ## [3.10.11] — 2026-04-28
 
 ### Fixed
--- a/public/app.html
+++ b/public/app.html
@@ -1594,7 +1594,13 @@
             </div>
           </div>
           <div class="shortcut-row" id="shortcut-row-ptt">
-            <span class="shortcut-label"><span data-i18n="settings.desktop_shortcuts_section.ptt">Push-to-Talk</span> <span class="muted-text" style="font-weight:400;font-size:0.85em" data-i18n="settings.desktop_shortcuts_section.ptt_toggle">(toggle)</span></span>
+            <span class="shortcut-label">
+              <span data-i18n="settings.desktop_shortcuts_section.ptt">Push-to-Talk</span>
+              <select id="ptt-mode-select" class="settings-select" style="font-weight:400;font-size:0.85em;margin-left:6px;padding:1px 4px" title="How the PTT key behaves">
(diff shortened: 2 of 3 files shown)


ancsemi:
Implemented and merged in 3.10.12. Lone modifiers (just Alt, Ctrl, Shift, etc.) now arm a pending state on keydown and commit on keyup if nothing else was pressed first -- great for gaming where you can't afford to move your hand. Mouse4/Mouse5 are now capturable too. The hardcoded "(toggle)" label is replaced with a hold/toggle select that gets saved through `setConfig`. OS-level registration of bare modifiers and extra mouse buttons is a separate follow-up in Haven-Desktop; for now, older desktop builds will get an explanatory toast instead of the generic "key is in use" error. Thanks for the PR!

ancsemi:
Hey, coming back to this one -- I want to be upfront about something I just realized. When I said "implemented and merged" earlier I was actually applying the changes manually and then closing the PR, which means you ended up with a closed (not merged) badge and no contribution credit on your profile. That's my mistake, not yours. I'm still figuring out the right GitHub workflow and only just realized that closing a PR after manually applying the code is basically the same as rejecting it from GitHub's perspective. Your work is in the codebase and has been since 3.10.12, but you deserved the merge credit for it and I dropped the ball on that. Sorry about that.

PR #5342, opened 2026-05-06
chore(i18n): update Polish translation and README

dvw1xx:
## Changes
- Updated the Polish (`pl`) translation
- Updated the README to mark the Polish translation as human translated

## Notes
The Polish translation is maintained manually rather than being AI-generated 😄


ancsemi:
Merged, thank you so much for keeping the Polish translation human-maintained. The AI-generated locales are useful as a starting point but they always read a bit off, so having an actual native speaker take care of pl is genuinely valuable to the project. Really appreciate you taking the time to do this properly. 🙏

PR #5420, opened 2026-06-15
Update ru.json

CUBEEEK:


ancsemi:
I just merged your full retranslation over in #5421, which overlaps with this one, so GitHub will show this as conflicting now. Two of the three wording fixes here aren't in that bigger PR though, so they're still worth landing. Whenever you get a chance, would you mind rebasing this onto the latest main? Once it's clean I'll merge it too so you get credit for both. Thanks again for the help with the Russian translation!

Issue #220, opened 2026-04-04
[BUG?] View all members


TGS-:
While not 100% sure this is a bug, I assume it is based on the permission toggle.

On the left hand menu next to the Themes, Activities and Thank you sits what appears to be a "View all members" button with the same power as the one the permission seems to be intended to give. Meaning anyone can view all members regardless of that permission setting.

PR #5428, opened 2026-06-27
Implement group DM support and rename functionality

averyfaulk:
# feat: Group DM support — private conversations with 3+ participants

Adds private group conversations (3+ participants) to Haven, similar to Discord group DMs. Existing 1-on-1 DMs remain unchanged and continue to use ECDH encryption.

## What this adds

- **👥 Create Group DM** — click the new 👥 button next to "Direct Messages" in the sidebar header
- **Auto-generated names** — show "user1, user2, user3 +2" based on participant names until a custom name is set
- **✏️ Rename Group** — right-click a group DM to set a custom name (persisted server-side, survives restarts)
- **👤 Add People** — right-click a group DM to add more participants after creation
- **🚪 Leave Group** — right-click → "Leave Group" to remove yourself
- **Live updates** — all members see additions/removals/renames in real time without refreshing
- **👥 Group icon in sidebar** — 👥 instead of @ for group DMs, with member count shown
- **No E2E for groups** — 1-on-1 DMs remain E2E encrypted; group DMs skip ECDH (doesn't scale to multi-party)
- **Persistence** — uses existing `channels.is_dm` + `channel_members` tables, no schema migration needed

### Server-side — `src/socketHandlers/`

| File | What changed |
|------|-------------|
| `channels.js` | **+172 lines** — added `start-group-dm` (creates DM with 3+ participants), `add-dm-member`, `remove-dm-member`, and `rename-dm` socket events. All validate bans, duplicates, and minimum participant count. |
| `index.js` | **+23 lines** — `getEnrichedChannels()` now detects group DMs (>2 members), populates `dm_members[]`, and auto-generates display names. Custom names are preserved on broadcastChannelLists() refresh. |

### Client-side — `public/js/modules/`

| File | What changed |
|------|-------------|
| `app-channels.js` | **+178 lines** — Group DM creation modal with user search + multi-select; context menu for Add People, Rename Group, Leave Group; 👥 icon + member count in sidebar; group name display in organize panel. |
| `app-socket.js` | **+54 lines** — `dm-members-updated` and `dm-left` socket handlers for live member changes; group DM name generation on `dm-opened`; E2E key fetch skips group DMs. |
| `app-utilities.js` | **+14 lines** — PiP popup header shows 👥 GroupName instead of @ DM for group conversations. |

### Client-side — `public/`

| File | What changed |
|------|-------------|
| `app.html` | **+23 lines** — 👥 create-group-dm button in DM sidebar header; Add People / Rename Group / Leave Group context menu items; Create Group DM modal HTML. |
| `css/style.css` | **+110 lines** — Group DM modal styles (user rows, checkboxes, hover states, disabled button). Note: some CSS is duplicated across media query blocks to match existing style patterns. |

### Other

| File | What changed |
|------|-------------|
| `package-lock.json` | Version bump (incidental — matches the fork base). |

## No schema changes

Group DMs reuse the existing `channels.is_dm` + `channel_members` tables. A channel is detected as a group DM at runtime when `channel_members` has >2 participants. The `is_group_dm` flag is computed in `getEnrichedChannels()` — no new columns or tables.

## Security & validation

- Guests cannot create group DMs
- Participants are validated against the bans table
- Duplicate membership is rejected
- Minimum 3 participants enforced server-side
- Channel codes validated via regex
- Name length capped at 50 characters
- Rename/remove operations check channel membership

## Testing notes

1. Start a new Haven server
2. Create 3+ user accounts
3. Click the 👥 button in the DM sidebar header
4. Select 2+ users, click "Create Group" — a group DM appears with 👥 icon
5. Right-click the group → Add People / Rename Group / Leave Group all work
6. Other members see changes in real time
7. 1-on-1 DMs continue to work as before with E2E encryption
8. Restart the server — custom group names persist, group DMs reappear correctly

## Rollback

All changes are contained in these files — revert individually or restore from git:

```bash
git revert 81d043e 3a2ac5f
```


ancsemi:
Thanks for putting this together, it's a ton of solid work and the socket flow reads really clean.

I hit one blocker before I can merge though. The create-group modal populates its user list from the `get-all-members` event, but that one is permission gated (server mods and up, or the `view_all_members` perm). It was locked down on purpose a while back in #220, so a regular member calling it gets "Permission denied" back and the modal just shows an empty list. That means normal users can't actually start a group, only admins/mods can.

Could you point that fetch at something a normal member is allowed to call? A small dedicated event that returns just the people they already share a channel or DM with would probably be the cleanest fit, rather than the full server member list.

Two smaller things while you're in there: `add-dm-member` and `remove-dm-member` only check `is_dm`, not that the channel is actually a group. So a crafted client could add a third person to a regular 1-on-1 (which is E2E encrypted and deliberately skips group encryption), or remove any member instead of just leaving. Worth gating those on the member count being over 2, and limiting removal to yourself.

No rush at all, happy to take another look once that part is sorted.

ancsemi:
Sat with this a bit more and I'm going to pass on it, so I'll close it here.

The sticking point for me is that group DMs skip the E2E encryption that regular DMs have (which is fair, multi-party key exchange is a whole different animal). But that means a group would look and feel exactly like a private DM while being a lot less private underneath, and I don't want to blur what the DM lock actually promises people. For a private conversation with a few people, a private channel already covers it without that tradeoff.

Genuinely appreciate the effort here though, the socket flow and the live member updates were nicely put together. This isn't a knock on the code, just a product call on my end.

PR #5442, opened 2026-07-19
feat: allow admins to reconfigure integration keys from Settings UI

KevonLin:
Previously the key setup form only rendered when a provider was not yet configured, so once STEAM_API_KEY (or Spotify/Last.fm keys) had been set, there was no way to replace them from the admin panel -- you had to manually edit .env and restart.

Changes:
- Render the setup form for admins regardless of configured state, hidden until revealed via "Set up" (unconfigured) or "Change key" (configured)
- Add a "Change key" button alongside Unlink/Link on every configured integration row, admin-only
- Wire click handler (connection-rekey) that toggles the form + auto-focus
- Adapt hint text: "this replaces the current key" when re-keying
- Trigger activity.pollSteam() immediately after STEAM_API_KEY is saved so the rotated key takes effect without waiting for the next 60s tick

Backend (envStore + set-integration-key) already supported re-writes; this closes the UI gap.

### DIFF
--- a/public/js/modules/app-users.js
+++ b/public/js/modules/app-users.js
@@ -418,30 +418,43 @@ _renderConnections() {
     const conn = linked.get(p.id);
     const configured = !!available[p.id];
 
-    let sub, btn;
+    let sub, btn = '';
     if (!configured) {
       // Admins get an inline setup form — most self-hosters have no idea where
       // .env lives, and telling them to "edit .env and restart" is a dead end.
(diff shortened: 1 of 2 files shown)


ancsemi:
Merged, thanks! Having to hand-edit .env just to swap a key was a real gap. Nice touch refreshing the Steam poll right after the save too, that's the kind of thing that's easy to miss and would have made the new key look broken for a minute.

PR #5443, opened 2026-07-19
feat: add PUBLIC_URL env var for OAuth/OpenID callback URLs

KevonLin:
When Haven runs behind Docker port mapping (8080->3000), a reverse proxy that strips the port from the Host header, or a Cloudflare Tunnel, the server can't reliably guess its own public-facing URL. This causes Steam and Spotify callback URLs to be wrong.

Add PUBLIC_URL environment variable. When set, baseUrl() uses it explicitly instead of trying to derive from req.protocol + Host. Covers all callback URL construction in connectRoutes.js (Steam OpenID return_to/realm and Spotify OAuth redirect_uri).

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -23,6 +23,13 @@ ADMIN_USERNAME=admin
 # Set to true to skip SSL even if certificates exist.
 # FORCE_HTTP=false
 
+# Public-facing URL for OAuth/OpenID callbacks.
+# Set this when the server cannot determine its own external address —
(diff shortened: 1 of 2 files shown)


ancsemi:
Merged, thank you! Docker port mapping and tunnel setups have tripped up a lot of people and this is a clean fix. Appreciate you keeping PUBLIC_URL out of the admin-writable keys as well, that's the safer call.

PR #5445, opened 2026-07-22
[feat] Auto-focus the message box when the user starts typing

Bo0sted:
Quality of life feature that mimics Discord.

When the user starts typing with no input focused and no popup or modal
open, the message textbox will automatically get focused so the user doesn't have to click into it.

Only plain printable keys trigger it: modifier combos (Ctrl/Meta/Alt) are left alone so shortcuts keep working, and it stays inert while another field is focused, during IME composition, or when the message area isn't visible. The handler never calls preventDefault, so the browser's own text insertion lands in the newly focused textarea and the first character is
never dropped.

### DIFF
--- a/public/js/modules/app-ui.js
+++ b/public/js/modules/app-ui.js
@@ -1740,6 +1740,24 @@ _setupUI() {
 
   // Global keyboard shortcuts
   document.addEventListener('keydown', (e) => {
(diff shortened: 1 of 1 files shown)


ancsemi:
Merged! This is a nice little quality of life touch, and I like that it stays out of the way during IME and keyboard shortcuts. Thanks Matt.

PR #5446, opened 2026-07-22
[feat] Add right-click context menu for messages

Bo0sted:
Quality of life feature that mimics Discord.

Mirrors the hover toolbar's actions (edit, reply, quote, pin, react, thread, copy-link, archive, delete) in a themed, cursor-positioned menu on the main message pane. Reuses the existing toolbar dispatch methods and channel-ctx-menu styling; gates each item on the same permissions. Only opens over a real message row, defers to the image menu on images and to native copy when text is selected.

### DIFF
--- a/public/js/modules/app-messages.js
+++ b/public/js/modules/app-messages.js
@@ -1800,4 +1800,136 @@ _initMoveMessages() {
   });
 },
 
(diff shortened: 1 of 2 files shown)


ancsemi:
This one is great. Mirroring the hover toolbar so closely means it picks up every theme for free, including win95. Merged, thank you.

PR #5447, opened 2026-07-23
docs: add STEAM_API_KEY to docs, add override.yml to .gitignore

KevonLin:
- README.md: add STEAM_API_KEY to config table with descriptions
- .env.example: add STEAM_API_KEY entry with usage docs and link to Steam dev portal
- .gitignore: add docker-compose.override.yml (local per-deployment config)


ancsemi:
Good call giving the Steam key a proper home in the docs and config table. Merged, thanks for this.

PR #5448, opened 2026-07-23
fix: resolve Steam OpenID callback failure behind CDN on non-standard origin port

KevonLin:
Root cause #1 — stale data-directory .env
  When Haven runs without Docker (or if the data-directory .env was created
  before PUBLIC_URL was added to .env.example), process.env.PUBLIC_URL is
  never set.  The server only loaded ENV_PATH (AppData/Haven/.env), so
  PUBLIC_URL from the project root .env was silently ignored.  Add a
  supplementary dotenv load of the root .env as a fallback, so manually
  configured env vars are always available regardless of Docker usage.

Root cause #2 — baseUrl() fell back to the raw Host header
  Behind a CDN → proxy → Docker chain, req.get('host') returns the proxy's
  internal address (e.g. localhost:8443) rather than the external domain.

  The new fallback logic:

  1. If Express trust proxy is enabled AND X-Forwarded-Host is present, use it verbatim — it carries the browser's original hostname and port, and is the most reliable source.

  2. Otherwise, use req.hostname (respects X-Forwarded-Host via trust proxy when available, giving the real public hostname) and only append a port from the raw Host header when it is NOT a common reverse-proxy internal port (3000, 3001, 8080, 8000, 80, 443).

  This handles all deployment patterns:
  - CDN + proxy + Docker (PUBLIC_URL set, X-Forwarded-Host present)
  - Direct non-standard port like :8443 with PUBLIC_URL unset
  - Cloudflare tunnel without X-Forwarded-Host
  - Multiple reverse-proxy hops (X-Forwarded-Host comma-separated)
  - IPv6 addresses ([::1]:port)

Improvement — debug diagnostics
  Log the exact openid.return_to, realm, and Steam's check_authentication
  response when verification fails, so the admin can see what URL was
  actually constructed vs what Steam expected.

Refs: #5443 (PUBLIC_URL)

### DIFF
--- a/server.js
+++ b/server.js
@@ -26,6 +26,21 @@ if (!fs.existsSync(ENV_PATH)) {
 }
 
 require('dotenv').config({ path: ENV_PATH });
+
+// Also load the project root .env as an override source.
+// Docker compose injects it via env_file, but when running directly on the
+// host the data-directory .env may be stale (created before PUBLIC_URL was
+// added), so the root .env serves as a fallback for env vars the server
+// administrator has explicitly set.
+//
+// This must be done *after* ENV_PATH so the data-dir .env takes precedence
+// for server-generated values (JWT_SECRET, VAPID keys).
+const rootEnv = path.join(__dirname, '.env');
+if (fs.existsSync(rootEnv)) {
+  require('dotenv').config({ path: rootEnv, override: false });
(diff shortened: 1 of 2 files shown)


ancsemi:
Really clear breakdown of the two root causes. Nice work tracking down the X-Forwarded-Host behavior behind the CDN. Merged, thank you.

PR #5452, opened 2026-07-24
[doc] Refine remote access documentation

Bo0sted:
Refined remote access documentation.

Informed users of the dangers of port forwarding, and provides an alternate, safer means of remote access,


ancsemi:
This is a genuinely useful addition, thank you Matt. The port forwarding risks are explained clearly and the Tailscale walkthrough is exactly the kind of safer option people should know about. I did a small pass on the heading levels afterward so the steps nest under each method, but the content is all yours. Merged!

Issue #5296, opened 2026-04-26
Auto complete features in threads and DM-PiP


birdcrazy:
Threads and DM-PiP don't suggest autocomplete options for mentions, emojis, or slash-commands.

It would be nice if all message input fields behaved the same, with autocomplete features.

ancsemi:
Wired `@mention`, `:emoji:`, and `/slash` autocomplete into both the thread reply input and the DM PiP input. The dropdowns reparent to whichever input is active so positioning still anchors above the right textarea, and arrow keys / Enter / Tab / Escape all behave the same as in the main message input. On `main`.

birdcrazy:
Works great, much appreciated!

PR #5454, opened 2026-07-24
[fix] :emoji autocomplete menu dying under specific circumstances

Bo0sted:
Steps to reproduce:
1. Edit an existing message, and type :emoji_name to open the emoji autosuggest menu. You do not have to select an emoji.
2. Cancel or save the edit,  either one triggers it.
3. Back in the main chatbox, typing :emoji_name to open the emoji menu throws a `dd is null` exception and the menu will not open anywhere.

Workaround before this fix: refresh the page.

The #emoji-dropdown autocomplete box is a single shared node that #5296 re-parents next to whichever input is active. Opening the suggestions inside an inline message-edit box parks that node inside the message; saving or cancelling the edit runs contentEl.innerHTML = originalHtml, which deletes it. From then on document.getElementById('emoji-dropdown') returns null, and _showEmojiDropdown threw on dd.parentElement — silently killing :emoji autocomplete everywhere (including the main composer) until a full page refresh.

The node holds no state (its contents are rebuilt on every open), so guard the lookup and recreate it when missing. Robust to any teardown path, not just the edit box.

### DIFF
--- a/public/js/modules/app-admin.js
+++ b/public/js/modules/app-admin.js
@@ -1951,7 +1951,19 @@ _checkEmojiTrigger(inputEl) {
 },
 
 _showEmojiDropdown(query) {
-  const dd = document.getElementById('emoji-dropdown');
+  // #emoji-dropdown is a single shared node re-parented next to the active
+  // input (#5296). Opening the suggestions inside an inline message-edit box
+  // parks it in that message; saving/cancelling the edit wipes the message's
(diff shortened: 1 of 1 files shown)


ancsemi:
Thanks Matt, this was really clean and the writeup made the root cause obvious. Good catch that the shared dropdown node gets wiped when the inline edit box tears down its HTML. Recreating it on demand is the right call. Merged!

PR #5455, opened 2026-07-26
[big feat] Add GB/GBC/GBA emulation with server-synced saves

Bo0sted:
# Add GB / GBC / GBA emulation with server-synced battery saves

Adds a Game Boy / Game Boy Color / Game Boy Advance emulator to the Activities panel, backed by an admin-managed ROM library and per-user battery saves that follow you across devices. Emulation runs client-side through EmulatorJS (loaded from its CDN, same way Flash uses Ruffle) — only the ROMs and `.sav` files ever touch the server.

---

## Breaking changes

Read these before pulling — self-hosters need to act on the first two.

### 1. Flash ROM directory renamed: `public/games/roms/` -> `public/games/flash_roms/`

Now that GBA ROMs also live under `public/games/`, keeping everything in a single `roms/` folder would mix Flash `.swf` files with Game Boy ROMs. Flash now lives in `flash_roms/`, GBA in `gba_roms/`.

**Impact:** if you already downloaded the Flash games, they're sitting in the old `roms/` folder and will show up as **Not installed**. Either move them over:

```bash
mv public/games/roms/*.swf public/games/flash_roms/
```

…or just re-run the in-app "Download Flash Games" button. The registry paths and `.gitignore` were updated to match. (The download manifest URLs still point at the frozen upstream `roms/` commit — that's intentional, don't "fix" them.)

### 2. Docker volumes: ROMs now need their own mounts

ROMs live under `/app/public/games`, **not** under `/data`, so the single `haven_data` volume never covered them. Uploaded GBA ROMs and downloaded Flash games would vanish on the next container update. Two named volumes were added:

```yaml
volumes:
  - haven_data:/data
  - haven_flash_roms:/app/public/games/flash_roms
  - haven_gba_roms:/app/public/games/gba_roms
```

If you run a hand-rolled `docker run`, add the two `-v` mounts (see the updated README). Without them the library is ephemeral.

### 3. CSP now allows `cdn.emulatorjs.org`

EmulatorJS pulls its emulator cores and wasm client-side from `https://cdn.emulatorjs.org`, so it was added to `script-src`, `style-src`, and `worker-src`, plus `blob:` in `connect-src` for the wasm/core blob fetches. If you front Haven with your own stricter CSP at a reverse proxy, allow that origin or GBA games won't load.

---

## Features

- **GB / GBC / GBA emulation** in the Activities panel — Gambatte for GB/GBC, mGBA for GBA, picked automatically from the file extension.
- **Admin-managed ROM library** — admins upload `.gb`/`.gbc`/`.gba` (32 MB cap), everyone else browses and plays. Filenames are sanitized and never overwritten (`pokemon.gba` -> `pokemon1.gba` on a clash).
- **Server-synced battery saves** — in-game saves are pulled at boot and pushed on every change, so your progress follows you to any device. Saves are per-user even though the library is shared.
- **Saves manager** — a modal to list, export (`.sav`), delete, and import saves. Flags "orphaned" saves whose ROM has left the library.
- **Offline resilience** — if the server goes unreachable mid-session, a persistent bar offers a one-click backup download and syncing retries with backoff. On tab close, any not-yet-synced save is sent with `navigator.sendBeacon` (which survives page unload); if you're offline it falls back to downloading the `.sav` automatically so a last-second save is never lost.
- **Tabbed Activities panel** — Flash / GBA / Web tabs replace the flat list.
- **Misc** — volume control, WebGL-disabled warning, a one-time "only battery saves sync" notice, win95 theme support, and full i18n across all 7 locales.

---

## How it works

Emulation itself is entirely client-side — `gba.html` sets the EmulatorJS globals and loads `loader.js` off the CDN, exactly like the Flash games do with Ruffle. The server never runs an emulator; it only stores ROMs (a plain directory that admins upload into, listed live by `/api/gba-roms`) and battery saves. Saves go in a new `game_saves` table as BLOBs rather than on disk — the files are tiny, they ride along in the DB backup, writes are transactional so a crash can't half-write one, and `ON DELETE CASCADE` cleans them up with the user just like high scores.

The tricky part is keeping saves in sync without ever clobbering good data. Every save is keyed by `SHA-256(filename + rom bytes)`, so two copies of the same game in the library keep independent saves and there's no chance of a collision picking up someone else's file. At boot the client hashes the ROM, pulls the user's save, and injects it into the emulator's SRAM before arming the sync — that ordering is what stops a fresh empty cartridge from overwriting a real save on the server. From then on it pushes on change (debounced ~1.5s, since one in-game save writes SRAM in a few stages), with a blank-SRAM guard so an all-`0x00`/`0xFF` buffer never goes up. The one genuinely ambiguous case is "the server has no save" — that could mean *deleted elsewhere* or *just never synced from this device*. To tell them apart the client keeps a per-device "synced at least once" marker and the server sends an explicit `X-Haven-Save: present|absent` header (with `no-store`), so a local copy is only wiped on an unambiguous `absent` from a device that had synced before — never on a 404, an auth error, or a dropped connection. Everything else (offline bar, backup download, `beforeunload` beacon) is just making sure that when sync *can't* happen, you still walk away with your `.sav`.


Bo0sted:
Fixing the merge conflict now

Bo0sted:
Done. PR now cleanly merges with [e065e9e](https://github.com/ancsemi/Haven/commit/e065e9e7e024f20214d2ada149236653dda29700)

PR #5459, opened 2026-07-28
[feat] Emoji menu shortcuts

Bo0sted:
The goal of this PR is to allow users to insert emojis into their messages quicker and more comfortably without having to take their hands off the keyboard. Discord's emoji menu already behaves this way, and since most Haven users are migrating from Discord, it makes perfect sense to implement these features similar to the way Discord does. 

## Main features
- Ctrl+E is now registered as a global hotkey and toggles the open state for the emoji menu
- Added a selection box for emojis that is controllable with arrow keys
- Upon selecting desired emoji, pressing Enter will insert that emoji and automatically close the emoji menu. Shift+Enter will insert the emoji but not close the emoji menu
- Escape now closes the emoji menu

## Also includes
- The first emoji is highlighted the moment the menu opens (and after every search or category change), so Enter works immediately without arrowing first
- Clicking a category tab moves the highlight to that category's first emoji, so keyboard and mouse navigation stay in sync
- The highlight box is theme-aware and uses each theme's own accent color, so it looks right in all of them

## Bug fixes
- Upon opening the emoji menu, the menu's scroll position is always reset to the top. Firefox was preserving the previous scroll position across reopens while Chromium wasn't, which left the view and the highlight on different categories
- Arrowing up to the very first emoji no longer tucks it under the sticky category header

## How it works
Under the hood, none of this reimplements how an emoji actually gets inserted. Pressing Enter on a highlighted emoji just fires that emoji's existing click handler, so the keyboard and mouse paths always stay in sync, and Ctrl+E reuses the same toggle the emoji button already calls. The arrow-key selection is a single highlight class that moves through the grid, with up/down worked out from the emojis' real on-screen positions so it stays correct across categories that wrap into different numbers of rows.

### DIFF
--- a/public/css/style.css
+++ b/public/css/style.css
@@ -6985,6 +6985,9 @@ details.sound-section:not([open]) .sound-section-label::before {
   max-height: 240px;
   padding: 2px;
   position: relative; /* offsetParent for category sections, so tab clicks can scroll to them */
+  /* Reserve the sticky .emoji-cat-header strip (26px) so keyboard scrollIntoView
+     lands emoji below the header instead of tucked underneath it. */
+  scroll-padding-top: 26px;
 }
 
 .emoji-cat-grid {
--- a/public/css/style.css
+++ b/public/css/style.css
@@ -7026,6 +7026,13 @@ details.sound-section:not([open]) .sound-section-label::before {
   transform: scale(1.2);
 }
 
+/* Keyboard selection highlight (arrow-key navigation in the picker) */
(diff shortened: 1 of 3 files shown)


ancsemi:
Merged. Really like that Enter just fires the emoji's existing click handler instead of duplicating the insert logic, that keeps the two paths from drifting apart later. The Firefox scroll-position catch was a nice bit of digging too.

Shift+Enter to keep the menu open is a good touch. Thanks for this one.

PR #5464, opened 2026-08-01
[feat] Rem interface zoom that scales the whole UI

Bo0sted:
## Rem interface zoom

Replaces the four fixed font-size tiers (Small/Normal/Large/XL) with a Zoom slider that scales the **entire** UI

### Why
The old font size buttons only touched enumerated text elements, so most of the UI (server rail, sidebars, icons, spacing) never scaled properly

### How
- A Zoom slider (70–150%) drives the root `font-size` via `--ui-scale`. Because everything is sized in `rem`, one root change rescales text, spacing, avatars, the server rail, sidebars, headers, and icons together. And the flex layout **reflows** instead of running off screen. 
- Converted the remaining fixedpx chrome to rem (layout width vars, server rail, sidebars, icon boxes, structural padding/gaps). Media query breakpoints, borders, shadows, and positioning offsets deliberately stay in px.
- Density stays its own independent axis (rem spacing tokens).
- Zoom is applied pre-paint in `theme-init.js` (no flash on load), persists to `localStorage`, and migrates anyone's old small/normal/large/x-large setting to the equivalent scale.

### Notes
- At 100% the UI is pixel-identical to before — this is non-destructive at the default.

### DIFF
--- a/public/app.html
+++ b/public/app.html
@@ -544,7 +544,7 @@
         <div id="sb-sidebar-resize-handle" class="sb-sidebar-resize-handle"></div>
         <div class="sb-sidebar-header">&#x1F3B5; Soundboard</div>
         <div class="sb-sidebar-body">
-          <input type="text" id="sb-sidebar-search" placeholder="Search sounds..." class="settings-text-input" style="width:100%;box-sizing:border-box;font-size:12px">
+          <input type="text" id="sb-sidebar-search" placeholder="Search sounds..." class="settings-text-input" style="width:100%;box-sizing:border-box;font-size:0.75rem">
           <div id="sb-sidebar-grid" class="soundboard-grid sidebar-mode"></div>
         </div>
       </aside>
@@ -559,7 +559,7 @@
         <div class="right-sidebar-voice" id="right-sidebar-voice">
           <div class="panel">
             <div class="voice-panel-header">
-              <h5 class="panel-title" style="margin-bottom:0">🔊 <span data-i18n="right_sidebar.voice_title">Voice</span></h5>
+              <h5 class="panel-title" style="margin-bottom:0"><span class="title-emoji">🔊 </span><span data-i18n="right_sidebar.voice_title">Voice</span></h5>
               <button id="mobile-right-close" class="mobile-panel-close mobile-panel-close-right" data-i18n-title="modals.common.close" data-i18n-aria-label="right_sidebar.close_panel" title="Close" aria-label="Close panel">&times;</button>
               <div class="voice-header-actions">
                 <button id="voice-mute-btn-header" class="voice-header-btn" data-i18n-title="voice.mute" title="Mute" style="display:none">🎙️</button>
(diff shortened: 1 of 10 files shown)


ancsemi:
The conversion work here is really solid. I checked all 2823 changed CSS lines with a script and 2655 of them are exact px/16 conversions with zero math errors, and you left the media query breakpoints in px like you said. That is a lot of careful work.

The thing holding me up is that it also carries a few changes to Layout Density that aren't really about zoom:

- Compact now hides message avatars completely, plus the channel `#` hashes and the panel title emoji. Before it just shrank avatars to 28px.
- Density no longer affects text size at all. Compact used to set message text to 13px and Spacious to 15px.
- `button` / `input` / `select` now inherit font-size, so any control without an explicit size gets bigger than it was. That is the one spot where 100% isn't quite pixel identical to before.

I'm not saying any of those are wrong, I can see the reasoning. I just want to judge them on their own rather than have them ride along with the rem work, since hiding avatars in Compact is something people will notice right away.

Could you split it into two PRs? The rem conversion plus the zoom slider on its own I'd merge pretty happily. Then the density rework as a second one we can talk through.

Heads up that #5459 went in and touched style.css, so this will need a rebase either way.

One tiny thing for whenever you rebase: the `env(safe-area-inset-*, 0)` fallbacks want to stay `0px`. A unitless zero is invalid inside calc() and max(), so the whole declaration gets dropped if the fallback ever fires. It doesn't bite today because browsers always define those variables, but it's a sharp edge to leave in.

Bo0sted:
You're 100% right. Layout density has nothing to do with zoom, so I stripped those changes out. Compact keeps its avatars, hashes and title emoji, and density drives text size again, exactly like before. I also dropped the font-size inherit on the form controls.

Fixed the env(safe-area-inset-*) fallbacks to 0px too, and rebased on top of main now that #5459 is in.

So this is just the rem conversion plus the zoom slider now. Thank you for reviewing.

ancsemi:
Merged, thank you. Stripping the density work out was the right call and it made this much easier to say yes to.

I re-ran the same check on the new diff: 2677 conversions, still zero math errors, and I confirmed no `[data-density]` rules are touched at all now, the form control `font-size: inherit` is gone, and the safe-area fallbacks are back to `0px`. Rebasing straight onto current main was appreciated too.

One thing I fixed in a follow-up rather than sending you round again for a single character. This line kept a unitless fallback:

`max-height: calc(100vh - var(--thread-footer-offset, 0) - 1.75rem);`

That one bites where the safe-area ones don't, because `--thread-footer-offset` is only defined under `[data-desktop-app]`. On the web client the fallback is what actually runs, so the calc goes invalid and the whole max-height gets dropped, and the PiP thread panel loses its clamp to the viewport. I checked it in a browser to be sure: `0px` computes to 692px, `0` computes to none. Easy one to miss since it looks identical to the env() cases.

Also worth knowing, some chrome I added while your PR was open (the status button, the new toggle sliders, a settings button grid) was still in px, so I converted that in a second commit. Not your problem at all, just bad timing on my end.

The zoom slider is lovely to use. Nice work.

PR #5471, opened 2026-08-02
Fix silent crash on launch (swallowed bind errors), allow Node 24, and add opt-in Braid themes + layout plugin

Amnibro:
## Silent crash on launch — root cause and fix

Since 3.33.0 the global `uncaughtException` "keep alive" handler swallows **async bind failures**: `server.listen` has no `error` listener, so `EADDRINUSE`/`EACCES` falls through to the keep-alive, gets one line in `crash.log`, and the process lives on **with no listening socket** — no banner, no exit code, nothing on screen. Worse, `Start Haven.bat`'s netstat readiness probe is satisfied by whatever stale process still holds the port, so it prints "Haven is LIVE" while the new server is a zombie. Every retry stacks another invisible zombie. To users this presents as *Haven silently crashing on launch* (several of us hit it after upgrading; my own `crash.log` had the swallowed `EADDRINUSE` twice).

**Fix:** an `error` handler on the main listener makes bind failures fatal and self-explanatory (which port, why, what to do) before `exit 1`; all other server errors keep today's keep-alive behaviour. The HTTP→HTTPS redirect listener (`PORT+1`) gets a warn-and-continue handler so it can't feed the same silent path.

**Also: Node 24 refusal.** `Start Haven.bat` hard-blocks Node ≥24 with "Haven requires Node 18-22" — but Node 24 is the current LTS that new users install, and better-sqlite3 12.10.0 ships working prebuilds for it (verified end-to-end on 24.14: install, boot, DB ops). The bat now verifies compatibility honestly — `node -e "require('better-sqlite3')"` after `npm install` — and only warns on ≥25. `engines` widened to `>=18 <25` (lockfile version also synced from stale 3.31.2).

Repro (before → after):
```
# terminal 1                      # terminal 2
node server.js                    node server.js
# ...HAVEN is running...          # before: prints startup lines, then hangs forever, port dead
                                  # after:  ❌ Haven could not start: port 3000 is already in use — ...
                                  #         exit code 1
```

## Braid themes + Braid Layout plugin (pure opt-ins, zero core-file edits)

Ports the Braid look from the Amni-Haven mobile app onto the rails this repo already ships:

- **`themes/braid.theme.css` / `braid-light.theme.css`** — the mobile BRAID / BRAID_LIGHT palettes (mint on slate / deep green on white) with the 1.6×-rounder shape language. Unpublished by default; admins opt in per server via **Settings → Custom Themes**, like any file theme.
- **`plugins/BraidLayout.plugin.js`** — a "two edges only" simplified layout: folds the server rail (including manage/sync) into a strip atop the sidebar, gives messages the full width, and tucks header extras into a kebab menu. **Every stock feature stays reachable** — the kebab proxies search/pins/gallery/channel-code/encryption/settings and adds People, Soundboard, Status bar, and app-download entries for the chrome it hides. A separate **variable-driven shape layer** (radii/blur/focus rings, no hardcoded colors) applies the Braid format on top of *any* active theme — Matrix keeps its rain and green, Nord stays Nord, just braided. `stop()` fully restores DOM, styles, listeners, and localStorage (verified through enable→disable→re-enable). The MutationObserver is childList-only and mutates only styles/attributes, so it can't observe its own writes. All sizes are rem so the interface Zoom slider scales Braid chrome with everything else; `prefers-reduced-motion` disables the motion layer.

## Validation

- `node --check` across all 58 repo JS files; `scripts/validate-locales.js` passes
- Fresh-server boot, admin registration, theme publish + picker injection, plugin enable/disable/re-enable — all exercised in a real browser on 3.39.0
- `Start Haven.bat` chain boots on Node 24.14 (`/api/health` verified)

### DIFF
--- a/Start Haven.bat
+++ b/Start Haven.bat
@@ -70,19 +70,18 @@ exit /b 1
 for /f "tokens=1 delims=v." %%v in ('node -v 2^>nul') do set "NODE_MAJOR=%%v"
 echo  [OK] Node.js found: & node -v
 
-:: Warn if Node major version is too new (native modules won't have prebuilts)
+:: Warn if Node major version is very new (native modules may lack prebuilts).
+:: Don't hard-refuse on a version number — Node 24 is the current LTS and
+:: better-sqlite3 ships prebuilts for it.  The real gate is the functional
+:: native-module load check after npm install below.
 if defined NODE_MAJOR (
-    if %NODE_MAJOR% GEQ 24 (
+    if %NODE_MAJOR% GEQ 25 (
         color 0E
         echo.
-        echo  [!] WARNING: Node.js v%NODE_MAJOR% detected. Haven requires Node 18-22.
-        echo      Native modules like better-sqlite3 may not have prebuilt
-        echo      binaries yet, causing build failures.
+        echo  [!] WARNING: Node.js v%NODE_MAJOR% detected. Haven is tested on Node 18-24.
+        echo      If the native module check fails below, install
+        echo      Node.js 24 LTS from https://nodejs.org
         echo.
-        echo      Please install Node.js 22 LTS from https://nodejs.org
-        echo.
-        pause
-        exit /b 1
     )
 )
 
@@ -94,6 +93,18 @@ if %ERRORLEVEL% NEQ 0 (
     color 0C
     echo.
     echo  [ERROR] npm install failed. Check the errors above.
+    echo.
+    pause
+    exit /b 1
+)
+:: Verify native modules actually load on this Node version (the honest
+:: compatibility test — version-number guessing refuses working setups).
+node -e "require('better-sqlite3')" >nul 2>&1
+if %ERRORLEVEL% NEQ 0 (
+    color 0C
+    echo.
+    echo  [ERROR] The better-sqlite3 native module failed to load on Node v%NODE_MAJOR%.
+    echo          Install Node.js 24 LTS from https://nodejs.org and try again.
     echo.
     pause
     exit /b 1
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "haven",
-  "version": "3.31.2",
+  "version": "3.39.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "haven",
-      "version": "3.31.2",
+      "version": "3.39.0",
       "license": "AGPL-3.0",
       "dependencies": {
         "adm-zip": "^0.5.16",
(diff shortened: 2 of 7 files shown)


Amnibro:
Pushed a follow-up commit (0b35753): merged message-run cards matching the mobile Braid shape language (O(n) run marking — no `:has()`, which went quadratic on long channels), own-message accent tint, carded channel runs, and condensed chrome (topic bar folds into the header band, tighter composer/sidebar). All rem-based so it tracks interface zoom.

Amnibro:
a726750 rounds out the plugin: full Mod Mode interop (the layer suspends while the layout editor is active so every section drag and panel snap works, then re-applies honouring the saved order), currentColor line icons for the bottom-left buttons, card-based settings with a pill nav rail, and subtler dotted seams inside message runs.

Amnibro:
14e8895: author-run message merging (one card per author regardless of the app's compact grouping, hover timestamps in the gutter), the header kebab + bottom-left icon bar collapse into one animated hamburger menu mounted on <body> (the blurred header clips even fixed-position descendants), continuous chat/UI text-size sliders in Settings, a breathing accent glow on the brand hexagon, a hexagonal channel-switch loader, and a type/color enforcement pass in both Braid themes.

Amnibro:
60878bc: the hamburger moves to the bottom-left corner (where the popups it opens anchor) as a Haven-styled tile with the same morph/stagger animation, and the settings nav becomes full-width pills — stock's hanging-indent emoji hack pulled row text left out of the highlight — with currentColor line glyphs replacing every leading emoji (nav rows + User/Admin tabs, hexagon fallback for unmapped rows).

ancsemi:
Merged. The launch bug is a good find and the writeup made it easy to check.

I reproduced the swallowed bind on its own before merging: with the keep-alive handler and no error listener, EADDRINUSE gets logged and the process stays alive with listening=false and exits 0. A zombie holding nothing, exactly as you described. With the listener it exits 1 with the reason. Nice catch on the redirect listener feeding the same path too.

The Node change I like even more, because the old bat hard-refused at 24 and exited, which meant it would have refused to start on my own machine. Swapping a version guess for an actual better-sqlite3 load check is the right call.

One follow-up on that: I run Node 26 here, so I pushed the ceiling to <27 and pointed the messages at 26. Your functional check is doing the real work either way, the number just needed to stop being wrong.

Read the Braid plugin and themes in full before merging. Restore looks solid (previous localStorage values recorded and put back, listeners tracked, braidOrig on anything you replace) and nothing in there touches usernames or message content, so no injection surface. Both stay opt-in, which is the important part. The kebab keeping every hidden control reachable is a thoughtful touch.

PR #5472, opened 2026-08-02
GIF search: add Tenor as the primary provider (GIPHY no longer issues API keys)

Amnibro:
## Problem

GIF search is dead for new servers: the picker's setup guide walks admins to developers.giphy.com, but **GIPHY no longer issues API keys to new applications** — the flow dead-ends and `/api/gif/*` stays at `gif_not_configured` forever.

## Fix

Add **Tenor (v2)** as the preferred provider; existing GIPHY keys keep working untouched.

- `getGifProvider()` resolves `tenor_api_key` (admin setting or `TENOR_API_KEY` env) first, then the legacy GIPHY key. Both providers normalize to the `{ id, title, tiny, full }` shape the client already consumes — the picker, favorites, and slash-command grid need no changes.
- Responses carry `provider`, and the picker footer reports whoever actually served the batch ("Powered by Tenor" / "Powered by GIPHY").
- The in-picker admin setup guide shows Tenor steps by default (Google Cloud console → enable Tenor API → API key) with a provider selector for admins who still hold a GIPHY key.
- `tenor_api_key` joins the settings allowlist, the sensitive-key redaction list, and gets the same validation as the GIPHY key.
- `media.tenor.com` / `c.tenor.com` join the giphy hosts in the inline-image whitelists so sent GIFs render as images.
- `.env.example` and strings updated; `validate-locales` passes.

## Verified against a live server

| Config | Result |
|---|---|
| `TENOR_API_KEY` set | `{"provider":"tenor","results":[...]}` |
| legacy GIPHY key only | `{"provider":"giphy","results":[...]}` |
| no key | `501 {"error":"gif_not_configured"}` → admin setup guide |

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -57,7 +57,10 @@ ADMIN_USERNAME=admin
 # TURN_USERNAME=haven
 # TURN_PASSWORD=your-password
 
-# Optional: GIPHY API key (free — get one at https://developers.giphy.com/dashboard/)
+# Optional: Tenor API key for GIF search (free — enable the Tenor API and create
+# an API key at https://console.cloud.google.com/)
+# TENOR_API_KEY=
+# Optional: GIPHY API key (legacy — GIPHY no longer issues keys to new apps)
 # GIPHY_API_KEY=
 
 # ── Dynamic DNS auto-updater ─────────────────────────────
--- a/public/app.html
+++ b/public/app.html
@@ -515,7 +515,7 @@
                 <button type="button" class="gif-tab" data-gif-tab="favorites" data-i18n="gifs.tab_favorites">★ Favorites</button>
               </div>
               <div class="gif-picker-header">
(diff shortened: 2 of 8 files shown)


ancsemi:
Merged, and thanks for spotting that GIPHY had quietly closed the door on new keys. That would have looked like a broken feature to every new admin with no way to tell why.

Normalizing both providers to the same shape was the right move, it kept the whole client side untouched. Checked the parts I care about most: the key never leaves the server, tenor_api_key made it into the sensitive list so non-admins can't read it back, and the host patterns are anchored properly so nothing can slip past them with a lookalike domain. Also nice that swapping to a parameterized lookup tidied the old string-built query.

Worth noting for anyone reading later: contentfilter=off on Tenor is not "no filter at all", it lines up with the r rating GIPHY was already using, so this isn't a loosening.

birdcrazy:
I'm confused. It looks like Tenor is also no longer accepting new API keys?
https://developers.google.com/tenor/guides/quickstart

image
-
-
image

PR #5475, opened 2026-08-03
[feat] Referrer-Policy header setting

Bo0sted:
## Referrer-Policy header setting
Adds a new Admin menu category "Security", and a new sub-category "Referrer configuration" to manage what Referrer-Policy should apply to all users. This new sub-category contains all eight referrer policies, but only documents three I believe will be the most useful to users while leaving the other options deliberately undocumented. This change inherits the old, hardcoded `strict-origin-when-cross-origin` so everything stays the same unless an admin explicitly changes it after this update.

### Why
Haven hardcoded the Referrer-Policy to strict-origin-when-cross-origin. That sends the site origin on cross-origin requests, and some CDNs reject it like Twitter/X.com. While post information can get through, videos do not.

### How
The value lives in server_settings, cached in memory and applied per response by the security headers middleware (helmet no longer owns the header). It's read at boot and updated the moment it changes, so no restart. Default Referrer-Policy is configured to be the current default, which is strict-origin-when-cross-origin.

The value is validated on the server against the fixed list of eight tokens, so an admin can't set an arbitrary header value. Clients pick up the new policy on their next page load, since it's a response header.

### Testing
Toggled through the policies and confirmed the Referrer-Policy response header changes to match. Verified X/Twitter video plays under Same-Origin and No-Referrer, and 403s under the default.

### DIFF
--- a/public/app.html
+++ b/public/app.html
@@ -1156,6 +1156,7 @@
           <div class="settings-nav-item settings-nav-admin" data-target="section-update" style="display:none">⬆️ <span data-i18n="settings.nav.server_updates">Server Updates</span></div>
           <div class="settings-nav-item settings-nav-admin" data-target="section-branding" style="display:none">🏠 <span data-i18n="settings.nav.branding">Branding</span></div>
           <div class="settings-nav-item settings-nav-admin" data-target="section-members" style="display:none">👥 <span data-i18n="settings.nav.members">Members</span></div>
+          <div class="settings-nav-item settings-nav-admin" data-target="section-security" style="display:none">🔒 <span data-i18n="settings.nav.security">Security</span></div>
           <div class="settings-nav-item settings-nav-admin" data-target="section-whitelist" style="display:none">🛡️ <span data-i18n="settings.nav.whitelist">Whitelist</span></div>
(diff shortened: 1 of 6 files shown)


ancsemi:
Merged. The plumbing here is clean, handing the header off from helmet to the middleware and caching it in memory so there's no per-request DB read is exactly how I'd have wanted it done, and I appreciate that you kept the current value as the default so nothing changes for anyone who doesn't touch it.

One change I made on top, and I want to explain rather than just quietly do it. I dropped Unsafe-URL and No-Referrer-When-Downgrade from the list, so six of the eight remain. Both of those send the full page address to other sites, and Haven happens to keep secrets in that address: invite links come in as ?invite=CODE, and they only get cleaned out of the URL once the socket connects. So on that first paint, any image hosted somewhere else in the channel would hand the invite code to whoever hosts it. That's not something an admin would expect from a dropdown.

Nothing you were after is lost, since Same-Origin and No-Referrer both send nothing cross-origin, which is what fixes the X video case you were chasing.

Thanks for this one, it's a genuinely useful setting and the Twitter/X finding was a good catch.

PR #5476, opened 2026-08-03
Keep file themes from inheriting the previous theme's effects and CSS

Amnibro:
Two bugs in how `.theme.css` file themes are applied, both hit while switching between built-in themes and file themes. One file changed, `public/js/plugin-loader.js`.

## 1. Effects survived the switch

Every built-in theme button re-runs `applyEffects(_getStoredEffectMode())` after it switches (`theme.js`, `initThemeSwitcher`), so picking a new theme tears down the previous theme's overlays. The theme buttons injected for published file themes go through `applyFileTheme()` instead, which never touched the effect layer — so the FFX water and wave, the Matrix rain, the Nord snowfall and the rest kept running on top of the file theme.

Repro on `main`: pick **FFX**, then pick any published file theme.

```js
localStorage.haven_theme        // "file:mytheme.theme.css"
[...document.querySelectorAll('[id^="fx-"]')].map(e => e.id)
// ["fx-ffx-water", "fx-layers", "fx-ffx-wave"]   ← still there
```

With a light file theme this is very visible: a blue wash down the sidebar and a wave band across the bottom of a white UI.

`applyFileTheme()` now re-applies the stored effect mode the same way the built-in buttons do, plus `showEffectEditorIfDynamic()` so the per-effect sliders stay in sync. Both calls are `typeof`-guarded, since `plugin-loader.js` does not depend on `theme.js` being present.

## 2. An enabled published theme overrode the selected one

Settings → Themes lists **every** `.theme.css` with an on/off toggle, published ones included, and toggling one on adds it to `haven_enabled_themes`. `applyFileTheme()` ended with `reapplyEnabledThemes()`, which re-appends every enabled file *after* the theme that was just selected — so an enabled published theme won the cascade over whatever the picker had active.

Repro on `main`, with a dark theme and a light theme both published:

```js
HavenPluginLoader.enableTheme('mytheme-dark.theme.css');   // Settings → Themes toggle
document.querySelector('.theme-btn[data-theme="file:mytheme-light.theme.css"]').click();

[...document.querySelectorAll('link[id^="haven-theme"]')].map(l => l.id)
// ["haven-theme-mytheme-light.theme.css", "haven-theme-mytheme-dark.theme.css"]
getComputedStyle(document.documentElement).getPropertyValue('--bg-primary')
// the dark theme's value — selecting the light theme changed nothing on screen
```

It persists across reloads too, since `loadTheme()` re-injects from the same list on init.

Published themes are what the picker offers, so they are now treated as mutually exclusive selections rather than stackable overlays:

- `reapplyEnabledThemes(activeFile)` skips the active theme and any *other* published theme, so it does what its comment always said: re-inject the unpublished CSS tweaks on top of the active theme.
- Selecting a published theme drops the others from the enabled list, and `loadTheme()` only injects a published theme when it is the selected one, so a stale entry can't resurface after a reload.
- Toggling a published theme on in Settings selects it, and toggling the active one off falls back to the built-in default rather than leaving the picker pointing at a stylesheet that is no longer loaded — the toggle list and the picker can no longer disagree.

Unpublished themes are untouched and still stack on top of whatever theme is selected.

## Verification

Run against a local server with a fresh `HAVEN_DATA_DIR`, two published themes (one dark, one light) and one unpublished tweak theme:

| case | before | after |
|---|---|---|
| FFX → file theme | `fx-ffx-water`, `fx-ffx-wave` still mounted | effect layers torn down |
| dark published theme enabled in Settings, then light one picked | `--bg-primary` = dark theme's | `--bg-primary` = light theme's |
| same, after reload | still dark | still light |
| unpublished tweak + file theme | works | works, tweak still last in the cascade |
| file theme → built-in FFX | works | works, FFX effects re-arm |
| Settings toggle states | active theme shown unchecked | active theme checked, the other published one unchecked |

Also `node --check` over all 58 tracked JS files and `scripts/validate-locales.js` (6/6 pass).

### DIFF
--- a/public/js/plugin-loader.js
+++ b/public/js/plugin-loader.js
@@ -175,6 +175,16 @@ window.HavenPluginLoader = (function () {
   function setEnabledThemes(list) {
     localStorage.setItem('haven_enabled_themes', JSON.stringify(list));
   }
+  // Published themes are what the theme picker offers, so only one can be
+  // active at a time. Unpublished ones are additive CSS tweaks that stack on
+  // top of whatever theme is selected.
+  function isPublishedTheme(file) {
+    return !!loadedThemes.get(file)?.meta?.published;
+  }
+  function getActiveFileTheme() {
+    const saved = localStorage.getItem('haven_theme') || '';
+    return saved.startsWith('file:') ? saved.slice(5) : null;
+  }
 
   // ── Load a single plugin ──
   async function loadPlugin(meta) {
@@ -255,7 +265,11 @@ window.HavenPluginLoader = (function () {
   // ── Load a theme ──
   function loadTheme(meta) {
     if (loadedThemes.has(meta.file)) return;
-    const enabled = getEnabledThemes().includes(meta.file);
+    // A published theme is only injected when it is the selected one; leaving a
+    // previously-selected one in the enabled list must not stack it on top of
+    // whatever theme is active now.
+    const enabled = getEnabledThemes().includes(meta.file)
+      && (!meta.published || getActiveFileTheme() === meta.file);
     let linkEl = null;
     if (enabled) {
       linkEl = document.createElement('link');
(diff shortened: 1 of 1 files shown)


ancsemi:
Merged. The writeup with the repro snippets made this easy to follow, and the second bug in particular was one I'd have spent a while staring at, since it looks like the picker is just ignoring your click. Treating published themes as picker choices rather than stackable overlays is the right call. Thanks for chasing both.

PR #5477, opened 2026-08-05
Braid Layout v1.5: multi-server strip fix, in-sidebar voice dock, one-click layout toggle

Amnibro:
Follow-up polish round for the Braid Layout plugin from #5471, driven by daily-driving feedback. One file changed (`plugins/BraidLayout.plugin.js`), still fully opt-in and reversible.

## What's fixed

**Server strip broke with multiple servers.** `#server-list` is a block div, so inside Braid's horizontal strip the remote-server icons stacked vertically. The strip now flexes it inline — any number of servers rides one row, with a vertical separator, compact add/manage/sync buttons, and no visible scrollbar. Mod Mode's panel drag handles are excluded from the fold (previously one leaked into the strip per edit session, hidden from the editor's own cleanup).

**Voice calls were missing their controls.** Stock keeps every in-call control — camera, screen share, soundboard, listen-together, and the settings panel with the screen-share resolution/fps pickers — inside the right sidebar, which Braid collapses. The plugin now relocates the real `#voice-panel` and `#voice-settings-panel` (plus both mute/deafen button pairs) into a voice dock at the bottom of the left sidebar, with a People-panel toggle added. All voice JS resolves elements by id and the voice CSS is ancestor-free, so state syncing, titles, and quality settings work unchanged; original DOM positions are recorded and restored LIFO on suspend/disable.

**No easy way back to the classic layout.** The layout is now a one-click toggle in both directions, persisted per user: under Braid the bottom bar is [Menu] [Themes] [Classic layout]; in the classic layout a hexagon switch sits inside the stock bottom-left icon bar right next to the theme button. Ctrl+Shift+B (and Ctrl+Alt+B, since browsers reserve the former) flips it from the keyboard. Disabling the plugin via Settings never overwrites the user's layout preference.

**Mod Mode interop.** While the layout editor is active Braid still steps aside so every drag handle is real — and a floating "✓ Done editing layout" pill now makes the exit obvious. Round-tripping the editor repeatedly leaves the DOM byte-stable (verified across cycles).

**Assorted feedback fixes:** join/create sections restored to the top of the sidebar; message cards got more padding; button fills now mix from the theme's text color so controls separate from the background in light and dark palettes alike; the kebab menu became a compact icon + short-label list; and the sidebar width rule no longer uses `!important`, so the stock resize handle (and its saved width) works under Braid again.

## Testing

Validated live against a local server via automated browser sessions: multi-server strip with 4 remote entries, in-call dock with settings panel open, engage/disengage/Mod-Mode/reload matrix, both Braid themes plus a non-Braid theme for the contrast tokens. `node --check` passes; `scripts/validate-locales.js` passes.


ancsemi:
Merged. The voice dock is the one I'm happiest about, since losing screen share and the quality pickers behind a collapsed panel made the layout hard to actually live in. Nice touch keeping the Mod Mode handles out of the strip too, that leak would have been miserable to track down later.

I pushed one tiny follow-up on top: a comment above _suspend still pointed at _evalOrderCSS, which this PR removes. Nothing functional.

PR #5478, opened 2026-08-05
start.sh: stop killing healthy servers when the readiness probe can't run (curl-less hosts)

Amnibro:
Fixes the startup failure reported on the community server by TGS (and reproduced here on a fresh setup): `./start.sh` boots the server, the frontend visibly works, and 15 seconds later the launcher prints **"[ERROR] Server failed to start after 15 seconds."** and the server dies — with nothing in the logs except a graceful termination.

## Root cause

The readiness loop probes `/api/health` with **curl only**. Minimal server images (and hosts re-imaged after an incident) often have no curl at all — so every probe exits 127 while the server is already up and serving, and at the 15-second mark the launcher **kills the healthy process** and blames the server. Desktop VMs ship curl, which is why the same release "works in a VM" — that asymmetry was the tell in the report.

The "terminated gracefully" line in the log is the launcher's own `kill`.

## What changed

1. **The probe now runs on node itself** — node is guaranteed present (it runs the server), so there is no external HTTP-client dependency. Tries https (self-signed accepted) then http.
2. **A failed probe never kills a live server.** If readiness can't be confirmed after the window but the process is up, the launcher says exactly that (with pointers at crash.log / PORT / firewall) and leaves it running. Killing was only ever correct when the process itself died.
3. **A dead server process is detected immediately** — the loop checks `kill -0` each second, so a real crash (port in use, broken install) reports in ~1s with a pointer at the real log, instead of 15 silent seconds followed by a misleading message.
4. Window widened 15s → 30s for slow first-boot migrations.
5. The Node version gate now matches `Start Haven.bat`: instead of hard-refusing Node ≥24, it functionally verifies `better-sqlite3` loads (with one `npm rebuild` attempt and clear guidance on failure). Node 24 LTS works fine where the prebuilt/rebuilt module loads.

## Testing

- **Fix case:** started via `start.sh` with curl **and** wget shimmed to fail, on Node 24: functional gate passes, node-based probe confirms readiness, LIVE banner, `/api/health` serving.
- **Control repro:** the previous script under the same conditions served `/api/health` at t≈9s and was then killed at t=15s with the reported error — byte-for-byte the community symptom.
- **Crash case:** a server that exits during startup is reported in about a second with exit 1 and the real reason preserved above the launcher message.
- `bash -n` clean.

### DIFF
--- a/start.sh
+++ b/start.sh
@@ -42,12 +42,11 @@ if [ "$NODE_VER" -lt 18 ]; then
     echo -e "${YELLOW}  [!] Node.js 18+ recommended. You have v${NODE_VER}.${NC}"
 fi
 
-if [ "$NODE_VER" -ge 24 ]; then
-    echo -e "${RED}  [!] WARNING: Node.js v${NODE_VER} detected. Haven requires Node 18-22.${NC}"
-    echo "  better-sqlite3 does not ship prebuilt binaries for Node 24+,"
-    echo "  so npm install will fail without C++ build tools."
-    echo "  Install Node 22 LTS: https://nodejs.org/"
-    exit 1
+if [ "$NODE_VER" -ge 25 ]; then
+    echo -e "${YELLOW}  [!] Node.js v${NODE_VER} is newer than Haven has been tested with.${NC}"
+    echo "  Continuing — native modules are verified functionally below."
+elif [ "$NODE_VER" -ge 24 ]; then
+    echo "  [*] Node.js v${NODE_VER} — verifying native modules load (below)."
 fi
 
 # ── Install dependencies ───────────────────────────────────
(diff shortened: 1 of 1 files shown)


ancsemi:
Merged, thank you. The killing-a-healthy-server part is the bit that stings, and probing with node instead of curl is obviously right once you say it out loud, since node is the one thing guaranteed to be there.

I pushed one small follow-up on top: the script warned at Node 25+ and suggested Node 22 LTS if the native module wouldn't load, while Start Haven.bat warns at 27+ and points at Node 26. Two launchers in the same repo giving opposite advice is worse than either one on its own, so I matched them up. Everything else went in as written.

PR #5479, opened 2026-08-05
Braid Layout v1.6: respect every theme's identity; contain the server logo in its tile

Amnibro:
Two follow-up fixes for the Braid Layout plugin merged in #5477, from daily-driving it with the built-in themes.

## 1. The layout wasn't respecting other themes

Two distinct causes:

**Card surfaces were derived from `--bg-hover`.** On most palettes hover is a subtle tint, so it looked fine — but on themes where hover is a full-contrast *selection* color (win95's classic navy), every message card and channel row painted as a permanent highlight and text went illegible. The card token is now mixed from `text-primary` over `bg-secondary` at 5%, which reads as a subtle card on every palette, light or dark; hover states follow the same recipe.

**Cosmetic shape rules used `!important`.** Braid's radii and decorative borders beat every `[data-theme=…]` rule, so e.g. win95 lost its square corners. All purely-cosmetic shape rules (messages, buttons, modals, menus, toasts, reactions, inputs, code chips) now use `:where()` with no `!important`: they still override the base stylesheet, but any theme-specific rule wins. Braid keeps its geometry on themes that don't care, and steps aside for the ones that do.

Also: glyphs sitting on an accent-colored tile in the server strip (the home hexagon, hovered tiles) now use `--bg-primary`, so they stay visible on terminal-glow themes where `accent == text-primary` (matrix / fallout / crt) — previously the home glyph was invisible there.

**Verified live under the layout on matrix, win95, daylight, and ffx** — cards readable everywhere, win95 keeps its squares and navy selection, matrix keeps its rain and gets a visible home glyph.

## 2. The server logo became the background of the server tiles

The admin-set server icon is injected into `#home-server` as a bare, classless `` — and no stylesheet constrains it, so it renders at natural size and sprawls behind the neighboring tiles (most visible in Braid's horizontal strip, but the stock vertical rail has the same bug). `style.css` now sizes any direct `img` child of `.server-icon` (100% × 100%, `object-fit: cover`, inherited radius) — one rule fixes both layouts. Verified with a 512×512 injected logo: renders contained at exactly tile size, strip metrics unchanged.

Diff is the plugin plus one small rule block in `public/css/style.css`.

### DIFF
--- a/plugins/BraidLayout.plugin.js
+++ b/plugins/BraidLayout.plugin.js
@@ -1,8 +1,8 @@
 /**
  * @name Braid Layout
- * @description Vastly simplified two-edge layout: folds the server rail into the sidebar, docks the full voice controls bottom-left, tucks header extras into a kebab menu, merges message runs into cards, and calms the chrome. One-key toggle (Ctrl+Shift+B) between Braid and the classic layout. Suspends itself while Mod Mode edits the layout. Pairs with the Braid / Braid Light themes.
+ * @description Vastly simplified two-edge layout: folds the server rail into the sidebar, docks the full voice controls bottom-left, tucks header extras into a kebab menu, merges message runs into cards, and calms the chrome. One-key toggle (Ctrl+Shift+B) between Braid and the classic layout. Suspends itself while Mod Mode edits the layout. Pairs with the Braid / Braid Light themes, and respects every other theme: cosmetic shape rules use :where() so any [data-theme] override wins.
  * @author Amnibro
- * @version 1.5
+ * @version 1.6
  */
 class BraidLayout {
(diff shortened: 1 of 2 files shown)


ancsemi:
Merged. Deriving the card color from bg-hover was one of those things that looks completely fine until it isn't, and win95 turning every message into a permanent selection block is a good demonstration of why.

The :where() approach is neat. It drops the specificity low enough that any [data-theme] rule wins, while still sitting above the base stylesheet, so you get to keep the shapes without stomping on themes that care about their own.

Good catch on the server logo too. That one affects the normal vertical rail as well, it was just less obvious there.

PR #5480, opened 2026-08-05
[feat+fix] Jump down behavior + GIF window Escape fix

Bo0sted:
## What this does

Fixes the "jump to bottom" button and adds Escape as a shortcut for it.

### The button bug

The messages DOM is limited to 100 messages at a time. Once you scroll up
far enough, the latest messages get trimmed out, so the "bottom" of the DOM
isn't the real bottom of the channel anymore. The button scrolls to
that fake bottom, which triggers a fetch of the next page. This means
if you've scrolled up hundreds of messages (like clicking a pinned message link), 
you'd need to press the jump button many times to get to the real bottom.

Now the button checks whether the latest messages are actually loaded. If
they've been trimmed (`_noMoreFuture === false`), it re-fetches from the
present using the same path switching channels already uses, so one click
lands you at the real latest message.

### Escape shortcut

Escape now jumps to the latest message too, while a channel is open. It runs
in the capture phase, ahead of the other Escape handlers, and backs off if
anything is open that Escape should close instead (modals, search, the emoji
or GIF picker, autocomplete dropdowns, the image lightbox, context menus). So
it only jumps when Escape has nothing else to do, and never closes a
popup and jumps at the same time.

### GIF picker

While wiring the guard I noticed the GIF picker never closed on Escape, unlike
the emoji picker. Escape now closes it too, for consistency.

Both the button and the shortcut go through one shared `_jumpToLatest` helper.

## Testing

- Scroll to the top of a busy channel (click an old, pinned message), click the button once, land at the
  newest message.
- Press Escape in a channel with nothing open, same result.
- Press Escape with a modal, search box, GIF picker, or autocomplete dropdown
  open and confirm it only closes that and does not jump.

### DIFF
--- a/public/js/modules/app-socket.js
+++ b/public/js/modules/app-socket.js
@@ -892,11 +892,7 @@ _setupSocketListeners() {
 
     // Jump-to-bottom click handler
     if (jumpBtn) {
-      jumpBtn.addEventListener('click', () => {
-        this._scrollToBottom(true);
-        this._coupledToBottom = true;
-        jumpBtn.classList.remove('visible');
-      });
+      jumpBtn.addEventListener('click', () => this._jumpToLatest());
     }
 
     this._historyDebounce = 0; // timestamp of last history request
--- a/public/js/modules/app-ui.js
+++ b/public/js/modules/app-ui.js
@@ -1801,8 +1801,34 @@ _setupUI() {
       // restore runs, and only when it's open so Escape can't open it.
       const emojiPicker = document.getElementById('emoji-picker');
       if (emojiPicker && emojiPicker.style.display === 'flex') this._toggleEmojiPicker();
+      // Close the GIF picker too, matching the emoji picker's Escape behavior.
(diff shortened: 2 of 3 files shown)


ancsemi:
Merged. The button bug is a good find, and reusing the same reset the own-message path already does is the right way to snap back to the present rather than inventing a second way to do it.

I pushed a follow-up tightening the Escape guard, because running in the capture phase means anything the guard doesn't know about gets jumped out from under. Four things were slipping through:

The big one is editing a message. Escape cancels an edit, but that handler sits on the textarea in the bubble phase, so the jump won and either scrolled away from the box being typed in or, on a trimmed window, re-rendered it out of existence. Scrolling up to edit an old message is exactly the case the feature is for, so those two were bound to collide.

The DM PiP, thread and pins panels weren't in the list either, so Escape while typing in any of them reloaded the main channel sitting behind it. And .context-menu wasn't matching anything, since Haven's is .channel-ctx-menu.

Everything else went in as written.

PR #5481, opened 2026-08-06
[feat+fix] Cosmetic layer for Admin and role permission nuke bug

Bo0sted:
## What this does

Two things: fixes a bug where editing a role could wipe its permissions, and adds a cosmetic customization layer for the Admin role.

## Fix: role permission nuke

When a role was edited, the server rebuilt its permissions by deleting all of them and re-adding the ones that were submitted. For a non-admin, the re-add only kept a permission if the user still held it, and it checked that by reading the database. The problem is that check ran after the delete. So when a user edited the same role that granted their own permissions, the delete had already removed them, the check found nothing, and every permission got nuked for every member of the role, not just the editor.

It now works out the final set of permissions before deleting anything, and keeps any permission the user cannot control.

While I was in there I also added the level guard that was missing. There was no level check on editing at all, so a non-admin with `manage_roles` could edit or relevel any role, including their own top role, which let them bump their own level. Now a non-admin can only edit or create roles below their level, and cannot raise a role to their own level or above. This is enforced on the server. Both role editors also go read-only for roles at or above the user's level, and admin-only permissions are highlighted with a note.

## Feat: cosmetic Admin role

The Admin role is not a real role, there is no row for it in the roles table, so there was never a way to customize it. The admin was permanently stuck with a red name they couldn't change.

Now an admin sees an Admin entry pinned to the top of the Roles modal, separated by a divider, and only they can see it. Clicking it opens a cosmetic-only editor for:

- Name
- Color
- Icon
- A Visibility toggle. When off, the Admin role is hidden on profiles and its color is not applied to the name anywhere, for everyone.

This is 100% cosmetic. `is_admin`, the level (100) and the permission set are never touched, so the admin keeps every power no matter what is set. It is stored as one JSON row in `server_settings`, with no schema change and no migration, and servers that never save it behave exactly as before.

## How it is wired

- One server helper reads the cosmetic setting with safe defaults, and both the main role display path and the profile card use it, so the member list, chat, voice and tooltips all follow it from one place.
- Two admin-only socket handlers read and write the setting. Saving re-emits online users so everything updates live.

## How to test

**Permission nuke and level guard**
1. As a non-admin with `manage_roles`, edit your own role and try to change its permissions or bump its level. It should be rejected instead of wiping the role.
2. Try to edit a role at or above your level. The editor should be read-only.
3. As admin, confirm you can still edit everything normally.

**Admin cosmetic**
1. As admin, open the Roles modal. Admin sits on top, above a divider.
2. Change the name, color and icon, then save. Confirm it updates on your profile, in the member list and in chat.
3. Turn Visibility off and save. Your admin name should lose its color and role badge everywhere, for everyone, while you stay admin.
4. Log in as a non-admin and confirm the Admin entry is not visible in their Roles view.


ancsemi:
Merged, and the permission wipe is a properly nasty one. Deleting first and then asking the database whether the editor still holds each permission, inside the same transaction where the delete already happened, means the answer is always no. So the role loses everything, for everyone in it, not just the person editing. Working out the final set before touching anything is the right shape.

Keeping permissions the editor doesn't control instead of dropping them is an improvement on the old behavior too. Someone editing a role shouldn't quietly strip powers from it just because they can't grant those powers themselves.

The level guard was worth adding. Being able to relevel your own top role is the kind of thing nobody notices until someone does.

One behavior change worth knowing about: a non-admin can no longer edit a role at exactly their own level, only below it. That matches how assigning roles and kicking already work, so it's consistent, but if anyone has two mods at the same level sharing a role between them, that stops working.

On the cosmetic Admin role, I traced everything that reads the role helpers before merging, since returning an empty list for a hidden admin looked like it could bite. It can't: every caller uses it for name and color only, and the permission checks work off is_admin directly without going near it. So visibility off is genuinely just cosmetic, which is what you said, but it seemed worth confirming rather than assuming.

PR #5482, opened 2026-08-07
[feat] Display Name moderation

Bo0sted:
Adds a new permission called "Manage Display names". 

Members who hold the permission get a new "Edit Display Name" toggle in the existing Edit Nickname modal. 

Checking the toggle changes the mode of the modal. While checked, the modal will now change the target's Display Name which is visible to the entire server. A theme-aware note is also displayed to inform moderators of the mode change. 

Unchecking the toggle restores the modal to its original functionality, and will change the target's private nickname, which is only visible to the user making the edit. 

The rename updates the member list silently without posting an in-channel rename message, and is still recorded in the audit log.

### DIFF
--- a/public/js/modules/app-admin.js
+++ b/public/js/modules/app-admin.js
@@ -5,7 +5,7 @@ const ALL_PERMS = [
   'rename_channel', 'rename_sub_channel', 'set_channel_topic', 'manage_sub_channels',
   'create_channel', 'create_temp_channel', 'upload_files', 'use_voice', 'use_tts', 'manage_webhooks', 'mention_everyone', 'view_history',
   'view_all_members', 'view_channel_members', 'manage_emojis', 'manage_stickers', 'manage_soundboard', 'manage_music_queue', 'promote_user',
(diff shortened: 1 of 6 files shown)


ancsemi:
Merged, with a follow-up on top covering three things.

The main one is that a moderated name now sticks. Before this the person could open Settings and rename themselves straight back, which made the permission decorative in the case you'd actually reach for it. There's a lock flag on the user now, set when you rename someone and cleared when you reset them to their username, so self-renaming is refused while it's on. Admins are never locked, so nobody can get stranded with a name they can't change.

Second, it respects rank now, the way kick, ban and mute already do. There was no level check at all, so anyone with the permission could rename the server owner. Now it only works on people below your own level.

Third, the same automod pass a normal self-rename gets. That one is parity rather than a real fix: automod only looks at links right now, and display names can't contain dots, so it won't catch anything today either way. But the two paths shouldn't disagree, and if name rules ever land they'll reach both.

Checked end to end against a running server: the rename sticks, the target's own attempt is refused, resetting them releases the lock, and renaming an admin or a same-level peer is rejected. Everything else went in as written, and the modal toggle is a nice touch.

PR #5490, opened 2026-08-10
[feat] Build the emoji picker from the full Unicode list

Bo0sted:
The picker's list was hand-curated in app.js and missing plenty of emoji. Instead of topping it up by hand, the server now builds the whole categorised list from Unicode's own emoji-test.txt and serves it, and the client renders the glyphs with the OS font so they match whatever platform the user is on.

On boot the server keeps a copy of emoji-test.txt in the data dir, fetched from the draft directory so it tracks the latest release. It refreshes when the copy is over a month old, and a committed 16.0 file is the fallback when there's no internet, so the server always has an emoji list. A single stat, at most one request, at most one write per boot.

Parsing keeps fully-qualified and minimally-qualified entries (deduped by VS16), drops unqualified and component rows, and strips skin-toned variants down to the base glyph. Hair emoji come through as their own base glyphs, so the existing skin-tone code layers a tone on top the way Discord does (person red hair plus a tone). The skin-tone modifier-base set is derived from the same file. The built-in list stays as an offline fallback, and the bundled flag images are kept since Unicode flags don't render on every OS.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -86,3 +86,8 @@ ADMIN_USERNAME=admin
 #
 # Common:
 #   DDNS_INTERVAL_MINUTES=5               (how often to re-check; default 5)
+
+# Checks unicode.org every month for a new emoji standard and automatically regenerates the built-in emojis in Haven based on that standard
+# If you choose to leave this feature disabled, Haven will default to Unicode standard 16.0 which is bundled with the server
+# Uncomment to enable this feature
+# UNICODE_EMOJI_AUTO_UPDATE=true
--- a/public/app.html
+++ b/public/app.html
@@ -2517,6 +2517,14 @@
           <div style="margin-top:8px;">
             <button class="btn-sm btn-accent" id="open-emoji-manager-btn">⚙️ <span data-i18n="settings.admin.manage_emojis_btn">Manage Emojis</span></button>
           </div>
+          <small class="settings-group-label" style="margin-top:14px;display:block">🔄 <span data-i18n="settings.admin.emoji_auto_update_group">Built-in emoji updates</span></small>
+          <small class="settings-hint" style="display:block" data-i18n="settings.admin.emoji_auto_update_desc">Allow Haven to check Unicode.org for new emoji standards. This check is only performed at server boot, and only when the current standard file on disk was created over a month ago. If a new standard is available, Haven will automatically download the emoji pack and regenerate all your built-in emojis based on the new standard. This does not affect your custom emojis in any way.</small>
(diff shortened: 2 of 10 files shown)


ancsemi:
Really like this one, the hand-curated list has been overdue for replacement and building it from the source of truth is the right way to do it. Two things before I merge.

The URL is pulling drafts. I checked both:

```
Public/draft/emoji/emoji-test.txt   -> Version 18.0, dated 2026-04-30
Public/emoji/latest/emoji-test.txt  -> Version 17.0, dated 2025-08-04
```

So `draft/` is newer, but it's Unicode's in-progress data rather than a published release. Running your parser against it gives 9 emoji that no shipped OS font can render yet, so they'd show up as empty boxes in the picker, and that number climbs as more gets drafted into 18.0 over the year. I think `https://unicode.org/Public/emoji/latest/emoji-test.txt` is what you want.

The other thing is the download itself. A lot of people run Haven on boxes that either have no outbound access or where they'd rather it didn't reach out on its own, so I'd like a way to turn it off. Something like `EMOJI_AUTO_UPDATE=false` in the env, following the same shape as `ALLOW_PRIVATE_PREVIEWS` in server.js, with a note in `.env.example`. When it's off it just serves the committed file, which your fallback path already handles.

Everything else checked out. The endpoint is token gated, the picker renders through textContent so there's no injection surface, and the committed emoji-test.txt is clean. The refresh logic being one stat and at most one request per boot is a nice touch.

Bo0sted:
Got it. Pushed a new commit just now. Please review and let me know if its good :D 

I also fixed the jumbo view for emojis with skin tone modifiers which wasn't part of your feedback but figured this would be a great time to fix it.

ancsemi:
Merged, thanks Matt. You went further than I asked and made it opt-in rather than just opt-out, which is the better call, and the env override on top means locked-down deployments can pin it either way without touching the DB. Refreshing immediately when the toggle flips instead of waiting for the next boot was a nice touch too.

The skin tone fix is a good one. I checked the premise and you're right, the modifiers are Emoji_Modifier and not Extended_Pictographic, so the old pattern left the tone character behind:

```
"👋🏽"  old leftover: "🏽"   new leftover: ""
"👩🏾‍💻"  old leftover: "🏾"   new leftover: ""
```

Which is exactly why a toned emoji never got jumbo sized. Glad you folded it in.

One small thing I picked up after merging and have already pushed: the five data-i18n keys on the new settings block weren't in en.json. Nothing broke visually, since the applier leaves the element alone when the key is missing and your inline English stayed on screen, but with nothing in en.json translators had no way to see the strings. Added them so they can be localized. Worth remembering for next time, any new data-i18n key needs a matching entry in en.json.

PR #5491, opened 2026-08-10
[fix] Misc QOL improvements

Bo0sted:
## Misc quality of life improvements

Four small, self-contained fixes to message editing, spoilers, and the persona picker.

### Prompt to delete a message when its edit is cleared
Clearing a message during an edit and confirming used to silently cancel the edit. Now it offers to delete the message, matching Discord. It reuses the existing confirm modal, so Enter confirms the deletion.

### Hide custom emoji inside an unrevealed spoiler
A text spoiler hides its content with `color: transparent`, which works for text and standard emoji but not custom emoji, since those render as an `` that color can't touch. The custom emoji showed through the spoiler. It's now hidden until the spoiler is revealed, keeping its layout box so the covered area stays the right size.

### Carry a link's spoiler onto its embed
A link wrapped in `||spoiler||` hid its own text but its embed card still showed the title, description, and image in full. The embed now inherits the spoiler: it renders blurred behind the same tag used for spoiler images, and a click reveals it. Detection reads the rendered DOM so it covers every way a link can be marked up, and both embed paths are wired (the generic rich card and YouTube), so it applies to all links that embed rather than one type.

### Hide code blocks inside an unrevealed spoiler
An inline codeblock and multi-line codeblock also had the same bug with spoilers not applying properly.

### Stop the persona picker and emoji autocomplete from colliding
Typing `::` to start a persona message also tripped the `:emoji` autocomplete, so the two dropdowns fought over the same input. Since no emoji shortcode starts with a colon, a leading `::` can only be a persona, so the emoji autocomplete now leaves that token alone. Also scaled the persona avatar in the picker, which had no styling and rendered at full size and cut off, cropping it to a small circle the way the avatar shows once the message is sent.

### Let the browser's spell-check menu show inside the message editor
The message-row right-click handler opened Haven's custom context menu
(reply / quote / pin / delete) over everything in #messages, including
the inline edit textarea — so right-clicking a misspelled word there hid
the browser's spell-check suggestions. Defer to the native menu when the
target is inside .edit-textarea, where none of the menu items apply.

### DIFF
--- a/public/css/style.css
+++ b/public/css/style.css
@@ -9510,6 +9510,13 @@ html:not([data-toolbaricons="emoji"]):not([data-toolbaricons="color"]) .tb-icon-
   user-select: auto;
 }
 
+/* Custom emoji render as <img>, which `color: transparent` can't hide, so
+   they'd otherwise show through an unrevealed text spoiler. Keep them laid
+   out (the blur box stays the right size) but invisible until revealed. */
+.spoiler:not(.revealed) .custom-emoji {
+  visibility: hidden;
+}
+
 /* Spoiler images (sender marked the image as a spoiler) — blurred behind a
    label until clicked. The inner .chat-image is the actual image. */
 .spoiler-media {
--- a/public/css/style.css
+++ b/public/css/style.css
@@ -9538,6 +9538,32 @@ html:not([data-toolbaricons="emoji"]):not([data-toolbaricons="color"]) .tb-icon-
   visibility: hidden;
 }
 
+/* Inline code sets its own color and background, so `color: transparent`
+   can't hide it and it would show through an unrevealed spoiler. It's a
+   single line, so the spoiler's own bar covers it — just hide the element
+   (keeping its layout box) until revealed. */
+.spoiler:not(.revealed) .inline-code {
+  visibility: hidden;
+}
+
+/* Fenced code blocks are tall block-level boxes, but .spoiler is an inline
+   span whose background only paints the first and last line — the middle of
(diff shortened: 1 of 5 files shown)


ancsemi:
All of these are merged, thanks Matt.

The spoiler ones stood out the most to me. Custom emoji and code blocks rendering straight through an unrevealed spoiler kind of defeats the whole point of it, and carrying the spoiler onto the embed card is something I should have handled when embeds went in. The :: collision is one I never would have thought to go looking for.

ancsemi:
Following up on the spell-check commit, I'd like to keep it. I think your first instinct was right and the reboot just moved you to a spot where the bug doesn't happen.

The reason it looks like a non-issue is that the main composer is fine, and always was. `#message-input` sits outside `#messages`, so that contextmenu listener never fires there and you get the native menu every time. It's only the inline edit box that's inside a message row.

There's a selection guard a few lines below that looks like it should already cover you:

```js
const sel = window.getSelection?.();
if (sel && !sel.isCollapsed && msgEl.contains(sel.anchorNode)) return;
```

It doesn't, because Chromium doesn't expose a textarea's internal selection to the document Selection API. I checked in a browser with a textarea set up the same way the edit box is:

```
textarea has selected: "teh"
window.getSelection().isCollapsed: true
-> guard does not defer, preventDefault() runs
```

So the word is selected inside the textarea, the Selection object still reports collapsed, the guard falls through and we suppress the native menu. Your `.edit-textarea` check is what actually stops it.

Leaving it in. Thanks for flagging it either way, I'd rather hear about it twice than not at all.

Issue #5486, opened 2026-08-10
[Feature Request] LDAP Authentication


jesjhoward:
After reading Ancsemi's post on OIDC authentication, I would like to suggest LDAP as an alternative. This should allow for both issues to be taken care of (SSO availability and encryption token generation) I believe.

ancsemi:
Haven has OIDC sign-in now (Settings, Security), which covers the SSO half of this, and it's confirmed working with Authentik. LDAP would be a second directory protocol to carry, and it has the same catch OIDC had: Haven never sees a password it can derive DM encryption keys from, so it would need the same separate encryption passphrase. Everything I know of that speaks LDAP (Authentik, Keycloak, Authelia, FreeIPA, Active Directory through ADFS or a proxy) also speaks OIDC, so that's the route. Closing; if your directory genuinely has no OIDC front, say which one and I'll reconsider.

PR #5494, opened 2026-08-13
Braid: three spacing levels, and dark ink on the Android badge

Amnibro:
Two things, both reported against the bundled Braid theme.

### Braid spacing — Compact / Cozy / Spacious

Braid shipped exactly one spacing. This adds a three-level control in **Settings → Braid spacing**, and a **Menu → Braid spacing** entry that cycles (it deliberately leaves the menu open, since you want to watch the change happen behind it).

**Cozy is byte-for-byte the spacing that shipped**, so the two new levels go in both directions rather than only tightening. The same eight messages measure 417 / 513 / 617 px.

The scale is a set of tokens on `html[data-braid-density]`, read by both the form and layout sheets, so message runs and channel rows tighten together along with the sidebar width. **Corner radii are deliberately excluded from the scale** — this is a density control, not a shape one, and the run cards measure an identical 14px (12px for channel rows) at all three levels.

`--braid-gutter` is now derived as `calc(inset + avatar + row-gap)` instead of being hardcoded. A continuation indents by exactly its leader's gutter, so a level that shrinks the avatar can no longer drift a run out of line with its own first message. Measured 0px drift at all three levels.

Persisted through `HavenApi.Data`; an unrecognised stored value falls back to cozy. `stop()` and `_disengage()` remove the attribute, the settings card, and the injected style.

### Android badge contrast

`--accent-text` (#5486) fixed the accent-filled controls, but the Android App badge was left at 1.78:1. It can't follow the token because it paints itself a fixed Android-green gradient in every theme rather than using `var(--accent)`, so it kept a hardcoded `color:#fff`. It and its dismiss button now use an ink measuring **9.65:1** at the light end of that gradient and **6.46:1** at the dark end.

### Verification

Checked on a scratch server with channels and message runs seeded through the socket API:

| level | drift | card radius | channel radius | 8 messages | sidebar |
|---|---|---|---|---|---|
| Compact | 0px | 14px | 12px | 417.5px | 16.25rem |
| Cozy | 0px | 14px | 12px | 513.5px | 17.5rem |
| Spacious | 0px | 14px | 12px | 617.5px | 18.5rem |

Braid Light is untouched — its `#00875a` accent with white measures 4.55:1 and passes as-is.

One bug caught during that pass and fixed here: the density buttons were originally styled with inline `style=` attributes, which outrank the plugin's own `.active` rule, so the selected level rendered identically to the unselected ones. Button styling moved into the injected stylesheet.


ancsemi:
This one reads well, and I checked the contrast numbers against the badge gradient rather than taking them on faith. They land where you said, so that part is good to go.

It needs a rebase before I can merge it though. #5497 just went in, and the only thing actually conflicting now is CHANGELOG.md, where the Unreleased block lands on the same spot 3.47.0 went. I tried the merge locally and the plugin file itself combines with your other PR cleanly, density and the Join/Create fix both intact, so there is nothing to untangle there.

One small thing while you are in it: the header still says 1.7, which #5497 already used. Worth going to 1.8.

ancsemi:
Checking back in on this one. Nothing has changed on my side about the work itself, the contrast numbers still hold and I still want it.

Re-ran the merge against main today. It is still CHANGELOG.md and nothing else, just now against the 4.0.0 block instead of 3.47.0, so the rebase is the same small job it was two weeks ago. `plugins/BraidLayout.plugin.js` and `public/css/style.css` both still auto-merge clean, so the density work and the badge fix come across untouched.

The header is still at 1.7 and main is also at 1.7, so that bump to 1.8 is still open.

No rush from me, just did not want it to look like it had been forgotten.

ancsemi:
Reviewed, and the badge contrast fix is going in as soon as it merges. 1.78:1 is well under the floor and the gradient makes it worse at the light end, so measuring both ends and quoting the numbers is the right way to argue it.

Same situation as #5546 though: the only conflict is CHANGELOG.md and it's mine, 4.2.0 landed where your entry sits. Rebase and I'll merge it.

ancsemi:
One more thing now that #5507 has landed, so the rebase is one pass rather than two.

That PR took the plugin header to 1.8, which is the bump I'd asked for here back on the 15th. So this one wants 1.9 rather than 1.8 when you come to it. `plugins/BraidLayout.plugin.js` should still combine cleanly otherwise, the two of you were touching different parts of it.

Amnibro:
Merged `origin/main`. Changelog Unreleased is back on top of 4.2.0. Plugin header is **1.9** (#5507 already took 1.8). Density work and the Android badge contrast are unchanged.

ancsemi:
Merged, thanks. Same as #5546, I brought the branch up to main on my side to clear the changelog conflict rather than ask for a third rebase. The spacing levels and the badge ink came across untouched.

Issue #5470, opened 2026-08-02
Feature Idea: Email Invites


birdcrazy:
I think it would be cool if we had a way to send nicely formatted HTML invite emails via SMTP .

Ideas:
- A GUI in Haven to enter the name and email of the invitee, the name of the inviter, possibly a custom message to include with the invite, and a list of channels to add them to upon registering.
  - Users with a new "invite users" permission (globally) would have access to this.
  - Email address, invitee name, inviter-name, and custom message: don't get stored long-term; Only long enough to send the email.
  - Could also allow users with "manage channel roles" permission (globally) to optionally select role(s) to automatically assign the user upon registering
- Custom HTML email templates/themes?
- Unique, one-time invite-token per email (if server requires an invite token)
  - Limited time invites? Invites only valid for a specified time period. (eg. 24h, 7d, 1M, etc)
- Email includes a button/link that directs them to the registration page of the server. 
  - Potentially auto-filling the invite-token in the appropriate field?
- User is added to the channels they were invited to.

ancsemi:
Splitting this one up, because a good chunk of it already exists and you may not have found it.

Settings > Server Invite Code > Manage Invite Links does most of the list. Each link is its own code with its own set of channels, a use limit (set it to 1 and it's single-use), an expiry, and an on/off switch. The link drops people straight onto the registration page and carries the code through login and signup for you, so there's no token to type. Custom slugs work too, so you can hand out /invite/book-club rather than eight hex characters. That covers the one-time token, the time limit, the channel list and the auto-fill.

The piece that genuinely wasn't there is who's allowed to do it. That was admin-only, which made "you can invite people" mean "you can run the server". There's an invite-users permission on main now that covers exactly that. Holders see and manage only the links they made, capped at 25, and they get the Invite Links block without the server code, the vanity link or the default-channel settings, since those are server config. Nothing changes for admins.

On the email half, I'm going to leave it for now, and I'd rather say why than let it sit. It means SMTP settings, storing credentials, and then deliverability, which is the part that goes wrong quietly. Mail from a self-hosted box with no SPF or DKIM tends to land in spam, and the failure looks identical to the invite never being sent, so people would be chasing invites that technically worked. Against a link you can paste into whatever you already talk to that person on, the win is that it's prettier, and I don't think that's worth putting a mail server in everyone's setup.

[...]

PR #5496, opened 2026-08-14
Add Invitation Buttons

birdcrazy:
This Pull Request adds two buttons that open up the `invite-links-modal`. 
One Button is under the user list, the other is in the "All Members" window.

The buttons are only displayed when the user is admin, or has the "Invite Users" permission globally.

These buttons make it more apparent to the user that they have the ability to create invites, as well as eliminating the need to open/have-access-to the admin settings to create invites.
 
Demo Vid:

https://github.com/user-attachments/assets/100c64c8-ceec-44cb-9045-d75337ccf818

Refs #5470

### DIFF
--- a/public/app.html
+++ b/public/app.html
@@ -591,7 +591,7 @@
         <!-- Invite users button pinned under user list -->
         <div class="panel sidebar-invite-panel" id="sidebar-invite-panel" style="display:none">
           <button class="panel btn-invite-popout" id="right-btn-invite-popout" data-i18n-title="app.actions.invite" title="Invite Users">
-            <svg width="16" height="16" viewBox="0 0 32 32" class="invite-icon">
+            <svg width="16" height="16" viewBox="0 0 32 32" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round">
               <circle cx="14" cy="11" r="6"/>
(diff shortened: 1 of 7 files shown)


birdcrazy:
Unwanted behavior found, someone could use dev-tools to make the "Invite Users" button(s) visible and open the `invite-links-modal`.

Link creation is still blocked when attempted. But it would be better to not show the modal at all if the user doesn't have the requisite permissions

birdcrazy:
The model now refuses to open if the user doesn't have the required permissions, even if they gain access to the buttons. A toast message with the warning that they don't have access to create invitation links is shown when unauthorized modal opening is attempted.

Button visibility and model permissions previously only allowed admin and invite_users permissions.
Now they allow admin, manage_server, and invite_users permissions.

ancsemi:
Merged, thanks. Making invites reachable without going through admin settings is the right call, and gating on the global permission rather than the plain one matches what the server actually enforces, so it lines up nicely with the permission work in #5470.

I pushed a small follow-up after merging: the two new buttons pointed at translation keys that were not in en.json yet. English still rendered fine because the applier leaves the markup alone when a key is missing, but no other language could pick them up. Nothing wrong with your side of it, just the piece that is easy to miss.

PR #5497, opened 2026-08-14
Braid Layout: restore Join / Create a Channel, and fix the dead Encryption entry

Amnibro:
Reported by **birdycrazy**: the Braid layout "seems to completely remove the ability to join or create channels", and the **Encryption** menu entry does nothing.

Both reproduce. Join and Create were not merely collapsed — **they could not be reopened at all.**

### Join / Create were unreachable

`_collapseJoinCreate()` added a Braid-owned `braid-collapsed` class to the *section*, on top of the stock `.collapsed` class on the *body*, and the stylesheet hid `.braid-collapsed .collapsible-section-body` with `!important`. Nothing ever removed that class, so the stock toggle kept flipping `.collapsed` underneath while the body stayed hidden.

Traced on a live server, clicking the section header:

| | body classes | section classes | computed display |
|---|---|---|---|
| initial | `collapsible-section-body` | `sidebar-section braid-collapsed` | `none` |
| after 1st click | `… collapsed` | `sidebar-section braid-collapsed` | `none` |
| after 2nd click | `collapsible-section-body` | `sidebar-section braid-collapsed` | `none` |
| `braid-collapsed` removed by hand | `collapsible-section-body` | `sidebar-section` | **`block`** |

The stock toggle works perfectly; the Braid class simply outranks it forever. Net effect: an admin could not create a channel and nobody could join one while the layout was on.

Visibility is now driven only by the stock class, which the collapse helper still sets — so the sections keep opening collapsed, and the header reopens them. Verified after the fix: Join expands to a usable code field and Join button, Create expands to the name field and create button.

### The Encryption entry could never have worked

It proxied a click to `#e2e-menu-btn`, whose handler only toggles `#e2e-dropdown`. That dropdown lives inside `#e2e-menu-wrapper`, which `app-channels.js` sets to `display:none` outside a DM — so in a normal channel the entry was present and did precisely nothing. Confirmed live: `e2e-menu-wrapper` computed `display:none`, `e2e-menu-btn` not visible, entry still listed in the menu.

Inside a DM it would have fared no better — this layout hides the header chrome the dropdown anchors to, and the original click keeps bubbling to the document handler that closes the dropdown again.

Proxying the three actions directly avoids all of it, since each opens its own modal and none depend on the dropbox being visible:

- Verify encryption
- Recover keys
- Reset keys

The menu is built once, so these are tagged and resolved at open time — shown where encryption applies, hidden elsewhere. Verified: hidden in a channel, `display:flex` once the wrapper is shown as a DM does, and a click on the entry reaches the real `#e2e-verify-btn` handler.

### Testing

Live server, Braid theme published and the plugin enabled, checked as both a non-admin member and an admin. `node --check` passes. No other `braid-collapsed` references remain in the plugin.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "haven",
-  "version": "3.41.0",
+  "version": "3.47.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "haven",
-      "version": "3.41.0",
+      "version": "3.47.0",
       "license": "AGPL-3.0",
       "dependencies": {
         "adm-zip": "^0.6.0",
--- a/plugins/BraidLayout.plugin.js
+++ b/plugins/BraidLayout.plugin.js
@@ -2,7 +2,7 @@
  * @name Braid Layout
  * @description Vastly simplified two-edge layout: folds the server rail into the sidebar, docks the full voice controls bottom-left, tucks header extras into a kebab menu, merges message runs into cards, and calms the chrome. One-key toggle (Ctrl+Shift+B) between Braid and the classic layout. Suspends itself while Mod Mode edits the layout. Pairs with the Braid / Braid Light themes, and respects every other theme: cosmetic shape rules use :where() so any [data-theme] override wins.
  * @author Amnibro
- * @version 1.6
+ * @version 1.7
  */
 class BraidLayout {
   start() {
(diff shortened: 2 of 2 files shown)


ancsemi:
Merged, thanks. Both of these traced exactly as you described. Having Join and Create sit there looking clickable but never opening is the worst version of that bug, since nobody would report it as anything other than "the layout removed channels."

Good call proxying the three encryption actions directly rather than the dropdown.

PR #5498, opened 2026-08-14
Design: end-to-end encrypted group DMs (with signed messages)

Amnibro:
A design doc only — no code, following the `docs/livekit-sfu-plan.md` convention. Group DMs do not exist yet (`start-dm` takes a single `targetUserId`), so this is greenfield.

Revised from the first draft: rather than noting the sender-authentication gap and accepting it, this does it properly.

### The shape

1. **Add an ECDSA P-256 signing identity key**, wrapped in the same password-derived blob and published through the same TOFU-pinned flow as the existing ECDH key.
2. **One AES-256 epoch key per group**, wrapped to each member over the pairwise ECDH channel that already exists, rotated on every membership change.
3. **Every message signed** over a digest binding channel, epoch, sender and ciphertext.

Confidentiality comes from the epoch key, authenticity from the signature. Separating them is what makes a single shared group key safe: every member can decrypt, but only a signing-key holder can *author*.

### This closes a gap that exists today

The pairwise DM key is symmetric and nothing is signed, so **a DM partner can already forge a message attributed to you** and the UI renders it normally — sender identity is just the `user_id` column the server writes. The same signature fixes 1:1 and group together, and the doc proposes shipping both on one schedule.

### Why not a per-message ratchet

The textbook answer is Signal's sender keys with a ratchet. That is right for Signal and wrong here, for a structural reason:

Signal keeps decrypted history locally and never re-fetches. Haven does the opposite — the server holds ciphertext permanently and `decryptMessages()` runs on **every** history fetch, including back-pagination. Forward secrecy requires deleting old message keys; if they are gone, re-fetched history cannot be decrypted.

So a ratchet here buys forward secrecy only by making scrollback permanently unreadable, or by forcing clients to persist every message key forever — which is more key material at rest than the single epoch key it replaced, and worse on both counts.

Forward secrecy is therefore **epoch-granular**, with rotation policy as the tuning knob (always on membership change, optionally on a schedule). There is an acceptance criterion specifically for the regression a ratchet would have caused: scrollback must still decrypt after several rotations.

### P-256 over Ed25519

Reach only. WebCrypto, Android JCA and CryptoKit all cover P-256 signing with no polyfill and no new dependency; Ed25519 in WebCrypto is still uneven across browser versions. This has to work identically across three independent client implementations.

### Honest gaps

- **No post-compromise security.** MLS is the only real answer and needs per-device identity, credentials and delivery-service ordering Haven lacks, plus a Rust/WASM stack across all three clients.
- **No metadata privacy.** The server always knows group membership and timing.

### One server rule worth singling out

An epoch publish must contain **exactly one row per current member**. Without it, a malicious member could publish an epoch that silently omits someone — locking them out while the UI still shows them as a participant.

### DIFF
--- a/docs/group-dm-e2e-plan.md
+++ b/docs/group-dm-e2e-plan.md
@@ -78,7 +78,7 @@ CREATE TABLE dm_group_keys (
 Envelope, versioned so 1:1 `v:2` traffic is untouched:
 
 ```json
-{ "v": 3, "e": 4, "iv": "…", "ct": "…", "sig": "…" }
+{ "v": 3, "e": 4, "prev": "…", "iv": "…", "ct": "…", "sig": "…" }
 ```
 
 ---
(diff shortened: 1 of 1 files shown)


ancsemi:
Merged. I went through this against the code rather than taking the citations on faith, and they all hold up, including the one that matters most: there really is no signature anywhere in `e2e.js`, so the forge gap in 1:1 DMs is live today rather than something groups would introduce.

The part that sold me is the section on why not to copy Signal's ratchet. Haven re-decrypts history from the server on every scroll back, so deleting old keys would take scrollback with it, and having that written down with an acceptance criterion attached is worth more than the rest of the doc put together. The reasoning against tying trust to MAC or IP is worth keeping around for its own sake too, since that suggestion comes up periodically.

Comments on the implementation are over on #5499. Nothing there changes the shape of this.

PR #5499, opened 2026-08-14
Group DM E2E: crypto core, key distribution, and tests

Amnibro:
Implements the design from #5498 far enough to prove the properties hold. **No UI yet** — this is the layer everything else sits on, and it is separable and reviewable on its own.

Builds on: #5498 (design).

### What is here

**`public/js/e2e-group.js`** — the primitives, deliberately free of storage, sockets and DOM so the same file runs in the browser and under `node --test`.

- Confidentiality: AES-256 epoch key, wrapped per member over the pairwise ECDH secret that already exists. No new key agreement.
- Authenticity: ECDSA P-256 signature over a digest binding channel, epoch, sender and the previous message.
- Verification runs **before** decryption, and a rejected message returns `plaintext: null` — so a caller that ignores `ok` still cannot render a forgery.

The signature is what makes a single shared group key safe: every member can decrypt, but only a signing-key holder can *author*.

**`src/socketHandlers/groupE2E.js`** — distributes blobs the server cannot open, while enforcing what no client can verify for itself:

- only a current member may publish an epoch
- epochs are strictly sequential (concurrent rotations conflict and retry)
- a member reads **only** their own wrapped blob
- an epoch must cover **exactly** the current membership

That last rule is the one worth reviewing hardest. Without it, a member could publish an epoch that silently omits someone — locking them out while the UI still lists them as a participant.

**Migrations**: `users.signing_key`, `channels.key_epoch`, and the `dm_group_keys` table.

### Tests — 28, via `npm test`

This adds the repo's first real test suite (`node --test`, no framework).

The crypto tests are written as **attacks**, not happy paths:

| Attack | Result |
|---|---|
| Member forges a message as another member | rejected |
| Genuine message re-attributed to another sender | rejected |
| Ciphertext tampered | rejected |
| Signature stripped | rejected |
| Replayed into another channel | rejected |
| Replayed into another epoch | rejected |
| Reading an epoch you were never given | rejected |
| Unwrapping a key not wrapped for you | rejected |

Plus two properties from the design: **scrollback still decrypts across several epoch rotations** (the regression a per-message ratchet would have caused), and **concurrent sends sharing a `prev` do not false-alarm**.

The server suite boots a real Haven on a scratch port and covers the structural rules, including the omitted-member rejection and a check that one member's key fetch does not leak anyone else's blob.

**I verified the suite is load-bearing** rather than assuming it: disabling the signature check fails exactly the five attack tests and nothing else. A suite that only walks the happy path would pass just as well with the security removed.

### Not in this PR

Group DM UI, the client wiring in `e2e.js`/`app-platform.js`, Android and iOS ports, and roster-gossip publication. The primitive (`rosterDigest`) and the transcript checker (`verifyChain`) are implemented and tested here so the UI work has something settled to build against.


ancsemi:
The crypto core here is careful, and the server tests are the good kind, actually booting a server and trying the attacks rather than asserting against mocks. Verifying before decrypting, with the `{ok, plaintext}` shape so that ignoring `ok` gets you null, is a nice touch.

Three things I need changed before this can go in.

`setup.js` is not part of this. It looks like leftover Braid debugging: port 3211, an admin account with a fixed password, a "braid repro seed" message, tokens printed to the console. Nothing malicious about it, it only talks to localhost, but a file called `setup.js` sitting in the repo root that creates an admin account is exactly the kind of thing a self-hoster runs by mistake. Please drop it from the branch.

`rewrap-group-key` lets any member overwrite any other member's blob. The `ON CONFLICT DO UPDATE` has no check that the recipient actually asked for one, so a member can walk the list and replace everybody's copy. Best case that locks someone out of history, worst case it hands one person a private version of the conversation. The comment says scoping to a single recipient stops a whole-epoch rewrite, but calling it once per member does exactly that. It wants an outstanding request from that recipient before it will write, and the plan's own note about respecting the pin should be enforced server side rather than left to the client.

`start-group-dm` has no consent step and no dedup. Anyone can build a group out of arbitrary user ids and the server joins those people to it. `start-dm` reuses an existing DM where this makes a fresh channel on every call, so repeated calls just pile up channels and notifications for the people named. The global flood limiter holds it to roughly six a second rather than unlimited, but the API goes live the moment this merges even though there is no UI yet.

Two small ones: `generateChannelCode` already exists in `auth.js` and does the same thing, and `start-dm` wraps its insert in a try/catch where this does not.

Then two things about the plan rather than this PR, both worth settling before a signing key gets written anywhere.

The backup blob cannot be repacked in place. `_wrap` stores `JSON.stringify` of the bare ECDH private JWK, and `_unwrap` hands the result straight to `_importPair`. Putting a second key in there turns that plaintext into a container, and an older client reading it gets a JWK with no `kty` and no `d`, throws, and that person loses their DMs until they update. Web, Desktop and Android all release separately, so there is always a window where someone is on the old one. A second wrapped blob in its own field sidesteps the whole thing and stays additive.

Enforcement also needs a cutover point. Every message sent before signing existed carries no signature, so "never silently accept an unsigned message" cannot apply to history without flagging years of perfectly legitimate DMs. Grandfather everything before the cutover and enforce forward from there.

Happy with the direction overall. The 1:1 forge gap is real and worth closing on its own, and I would rather see the signing key and the message signatures land as their own PR ahead of any group work.

ancsemi:
Bumping this so it does not look abandoned. Main is at 4.0.0 now and I have not moved on anything I said above, the three items still stand as written.

Two things worth knowing before you pick it back up.

The rebase got smaller rather than bigger. `package.json` and `src/database.js` both auto-merge now; the only conflict left is `src/socketHandlers/index.js`, and only because the require block shifted when Ferry landed. Your two lines there go back in the same places.

And the test situation changed under you, in a good way. Main has a `test/` directory now with eight files, arrived through other PRs, run with `node --test "test/*.test.js"`. So this is no longer the repo's first suite and your two files drop in beside the existing ones rather than establishing the pattern. `package.json` still has no test script though, so that part of your diff is still wanted.

My preference from last time is unchanged: the signing key and 1:1 message signatures as their own PR first. The forge gap it closes is worth having on its own merits, it is a much smaller thing to review, and it settles the backup-blob question before any group work depends on the answer.

ancsemi:
Read the crypto properly rather than skimming it, and it holds up.

The signature covering channel, epoch, sender, prev, iv and ct is what makes the shared epoch key safe, and length-prefixing the signing input means no combination of fields can collide with another. Verify before decrypt with `plaintext: null` on failure is the right shape, since a caller that ignores `ok` still can't render a forgery. Separate ECDSA identity from the ECDH key is correct rather than incidental. Zeroing the raw key after export is a nice touch.

Server side, the rule you flagged is enforced the way you describe: sorted deduped recipients compared against current membership, so an epoch that quietly omits someone is refused. Sequential epochs with the UNIQUE violation caught as a conflict, wrapped blobs capped, and `get-group-keys` scoped to `recipient_id = socket.user.id` so a member only ever reads their own. That last one is the sort of thing that gets missed and it isn't.

Two things before this can go in.

`setup.js` in the repo root doesn't belong to this PR. It registers `admin` and `birdy` against port 3211 with a hardcoded password, publishes `braid.theme.css` and prints tokens. It's Braid repro tooling that rode along on the wrong branch. Please drop it.

It also needs a rebase. Main moved: `package.json` now has a test script (`node --test --test-concurrency=1`, serial because several integration tests spawn real servers and enough at once was crashing one of them), and `src/socketHandlers/index.js` has changed underneath you.

On merging it with no UI attached, I'm inclined to take it. It's tested, it stays dormant until something calls it, and reviewing 986 lines of crypto on its own is far better than finding it buried inside a 3000 line feature later.

ancsemi:
Correcting my last comment, which was wrong in a way that matters.

I read the crypto fresh without re-reading what I'd already written on this PR back in August, and came away saying two things stand between it and merging. It's four. Both items from the 15th are still in the branch, and I've just gone and checked rather than assuming either way.

`rewrap-group-key` still writes on any member's say-so. `request-group-rewrap` broadcasts a request, but nothing ties the two together: the handler checks that the caller is a member and the recipient is a member, then `ON CONFLICT DO UPDATE` overwrites. A member can still walk the roster and replace everyone's copy. It wants an outstanding request from that specific recipient before it will write, enforced here rather than left to the client.

`start-group-dm` still has no consent step and no dedup. Anyone can assemble a group from arbitrary user ids and the server joins those people to it, with a fresh channel on every call.

That also means I was too generous about the server side as a whole. The epoch publish path is tight and I stand by every word about it: sorted deduped recipients against current membership, sequential epochs, blobs capped, reads scoped to your own row. The rewrap path sitting next to it has none of that, and describing the enforcement as thorough without separating the two was the actual error in my last comment.

On scope, my preference hasn't changed either, whatever the last comment implied. Signing key and 1:1 message signatures as their own PR first. It closes a real forge gap on its own merits, it is a fraction of this to review, and it settles the backup blob question before any group work depends on the answer.

Nothing in your branch got worse since August. My reading of it did. The bar is where it was on the 15th.

Amnibro:
Merged `origin/main`. The four items from the 15th:

- Dropped `setup.js`.
- `rewrap-group-key` now requires an outstanding `request-group-rewrap` from that recipient and consumes it on write. The wrapper has to send `recipientPublicKey` matching the TOFU pin, or the server emits `public-key-conflict`.
- `start-group-dm` invites rather than joining people. Same roster reuses the existing group. `accept-group-dm` / `decline-group-dm` for the named people. Insert is in a try/catch, codes come from `generateUniqueSharedCode`.
- `package.json` keeps main's serial test script; Ferry and GroupE2E both stay registered.

I have not split signing keys / 1:1 signatures into their own PR. Still happy to if that's the path you want first. Backup blob stays a second field, not a repack of the ECDH wrap, and unsigned history still needs a cutover.

ancsemi:
Merged. All four items from the 15th are in and hold up on a fresh read: the rewrap path needs the recipient's own outstanding request and consumes it on write, a group starts as an invite the named people accept, the insert is wrapped and the code comes from the shared generator, and setup.js is gone. The 28 tests run green here alongside the rest of the suite.

One hardening I added after merging, in b2ac4e9. start-dm's lookup for an existing DM only checked that both people were members of an is_dm channel, so once a group exists that contains the same two people, a plain DM between them could have resolved to the group. It now requires exactly two members.

The signing key and 1:1 message signatures as their own PR is still the piece I'd like to see before any group UI, so yes, please do that one next when you have time.

Issue #5492, opened 2026-08-11
Permission Restriction Gap: Move Channel to Sub-channel


birdcrazy:
Branching from #5467 

Users who don't have permissions to manage sub-channels within `channel_A` can get around this restriction by creating their own main channel (`channel_B`), then move `channel_B` to be a sub-channel under `channel_A`

ancsemi:
Confirmed, and fixed in the next build.

The move-channel path was checking a permission that the server-wide **Create Channels** permission satisfies for *any* parent. So someone with Create Channels could make a top-level channel and slide it under a channel they don't moderate, exactly the bypass you described.

Nesting a channel under a parent now requires channel-scoped authority over that parent specifically: **admin, or Manage Sub-channels on the destination parent (or the channel's current parent)**. Create Channels still governs promoting/demoting your own top-level structure, but it no longer counts as authority over a parent you don't manage. Reordering and categories were left as-is.

Thanks for chasing the edge cases here.

birdcrazy:
This commit was added to [claude/haven-issues-5409-5426-5427-92t9no](https://github.com/ancsemi/Haven/compare/claude/haven-issues-5409-5426-5427-92t9no) branch. The commit did not make it to the main branch

ancsemi:
Fixed and on main, via #5500 from @birdcrazy.

You had it exactly right. The move check was satisfied by holding `create_channel` anywhere, because a top-level channel has no parent and the permission lookup for "no parent" quietly turned into a server-wide question. Nesting under a parent now needs `manage_sub_channels` on that specific parent, and the same applies going the other way, so pulling a sub-channel out of a parent you do not manage is refused too.

Moving a channel you own into a parent you manage still works, and so does taking it back out. Only the case you reported is blocked.

PR #5500, opened 2026-08-15
Fix subchannel restrictions

birdcrazy:
Users could still move a subchannel from a channel, or to a channel where they don't have the manage-subchannel permissions; They could also promote these subchannels.

This PR fixes these problems by adding the permission gate for these actions. Also hides the options in the channel context menu when not permitted.

refs: #5492

### DIFF
--- a/public/js/modules/app-channels.js
+++ b/public/js/modules/app-channels.js
@@ -383,6 +383,7 @@ _openChannelCtxMenu(code, btnEl) {
   // used to show the entry everywhere once you held the permission anywhere.
   const ch = this.channels.find(c => c.code === code);
   const canManageSettings = isAdmin || !!(ch && ch.canManageSettings);
+  const canManageSubs = isAdmin || !!(ch && ch.canManageSubs);
   const isMod = isAdmin || this._canModerate();
   menu.querySelectorAll('.admin-only').forEach(el => {
     el.style.display = canManageChannels ? '' : 'none';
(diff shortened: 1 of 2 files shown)


ancsemi:
Merged, thanks. The bypass is real and you found the exact line: `_canManageSubsOf` ends in a `create_channel` fallback, and on a top-level channel the parent is null, so that call was really asking "can this person create channels anywhere", which any channel creator passes. With the check being an OR, that one `true` meant the destination was never looked at. Gating the promote path was a good catch too, since it had the same shape going the other way.

I pushed one follow-up on top of it. `_canManageSubsScoped` fell through to `isAdmin` when the parent was null, so with the new AND, moving any top-level channel under a parent became admin-only, including a parent you fully manage. That left the two halves of one job disagreeing: you could promote a channel out of your own parent and then not put it back.

Top level is nobody's to manage, so a null parent now passes freely and imposes no requirement of its own. Touching top level at all is still gated on `create_channel` higher up in the handler. Each end of a move costs the same permission on the same parent, whichever direction you are going.

Your AND is doing the real work either way, and it is the right call. I checked the matrix against a running server as a non-admin with `create_channel` plus `manage_sub_channels` on one parent: nesting under a parent they do not manage is refused, promoting out of their own parent works, and moving it back under that parent works.

I also pointed the menu and the destination picker at the same rule. The picker was offering every top-level channel as a target including ones the server would refuse, and Move to was keyed on the channel being moved rather than the parents involved, so it could hide a move that was actually allowed.

PR #5502, opened 2026-08-16
Eliminate channel context menu divider

birdcrazy:
This prevents the double divider that can occur when both "Rename Channel" and "Create Sub-channel" options are not available in the channel context menu.

Double divider
image

Fixed (single divider)
image

Divider still displayed when the options are available.
image

Related to: #5492


ancsemi:
Merged, thanks!

ancsemi:
More usefully: this is in and the fix is right. Nice that you put the check after both the mod-only reset and the two per-button permission blocks, so it reads the final visibility rather than racing them.

One tweak on top, no action needed from you. The divider was found with an exact class-attribute match, which stops matching the moment anyone adds or reorders a class on that hr, and the logic would then quietly stop working with nothing to show for it. Switched it to match by class.

Issue #5501, opened 2026-08-16
OIDC/SSO provider unreachable


RCCore:
I have one more problem with the OIDC/SSO. When I put a slash at the end, it does not stick it. In the logs, it shows that it is using it without the slash, which is causing this problem, even though is it there.

- <img width="126" height="12" alt="Image" src="https://github.com/user-attachments/assets/60f2d488-b0cb-49ee-b0cb-6f4698c82d03" />

- <img width="342" height="142" alt="Image" src="https://github.com/user-attachments/assets/09f0f7e8-53ce-4091-a0c0-2737336d603f" />

And this is from the container log: "[OIDC] provider_unreachable: Provider issuer mismatch: configured https://something.com/application/o/haven, document says https://something.com/application/o/haven/"

ancsemi:
This is fixed on main now. Haven was stripping the trailing slash when it read the setting back, so what you typed was stored fine but the check compared the version without it. Authentik publishes the issuer with the slash, hence the mismatch.

@birdcrazy caught it and sent #5503, which stops Haven rewriting what you enter. I added a bit on top so it also works if someone leaves the slash off, and so the ID token check uses the provider's own spelling instead of the configured one, which would have been the next thing to fail for you.

It isn't in a release yet. If you're able to run main and confirm it works against your setup, that would be good to know before it ships.

RCCore:
It is working now, thank you @birdcrazy and @ancsemi

Issue #12, opened 2026-02-14
[FEATURE] Add OIDC authentication


Lordingard:
Hello,

Here a new feature proposal : OIDC authentication (for create/login with Authentik account for exemple).
I'm super excited about this project, thanks for the work!

ancsemi:
Hey, appreciate the suggestion! OIDC would definitely be cool for more advanced setups with Authentik, Authelia, etc.

Right now Haven's auth is pretty simple on purpose — just username/password — since most people running it are hosting for a small group of friends and I want the setup to stay as easy as possible. Adding OIDC would be a bigger lift and add some complexity that most users wouldn't need.

That said, I'm not ruling it out for the future. If there's enough demand for it I could see adding it as an optional thing down the road. Keeping this open so I don't lose track of it. Thanks for the kind words about the project! 🙏

Amnibro:
OIDC's already on the roadmap (#18) so this is partly just \"how do we want to land it.\" Sketch of a clean implementation that doesn't fork the existing local-auth path:

**Library:** \`openid-client\` (the maintained Node OIDC client). Lighter than passport, handles discovery / PKCE / token refresh out of the box.

**New schema:**

\`\`\`sql
ALTER TABLE users ADD COLUMN oidc_subject TEXT;        -- the IDP's `sub` claim
ALTER TABLE users ADD COLUMN oidc_issuer TEXT;          -- so we can verify on subsequent logins
CREATE INDEX idx_users_oidc ON users(oidc_issuer, oidc_subject);
\`\`\`

A user can have *either* a `password_hash` (local) or `oidc_subject` (federated) or *both* (link an existing local account to OIDC). Link is the trickiest UX — proposal: on first OIDC login where `email` matches an existing local user, prompt \"Link to your existing Haven account?\" with a password challenge.

**New routes:**

- `GET /api/auth/oidc/start` — opens session-bound state + PKCE verifier in a short-lived cookie, redirects to IDP authorization URL.
- `GET /api/auth/oidc/callback` — validates state + PKCE, exchanges code for tokens, looks up by `(issuer, sub)`, creates a new user or signs in existing, mints a Haven JWT, redirects to `/`.
- `GET /api/auth/oidc/logout` — optional RP-initiated logout if the IDP advertises it.

**Settings:**

\`\`\`
oidc_enabled         (default false)
oidc_issuer_url      (e.g. https://auth.example.com)
oidc_client_id
[...]

PR #5503, opened 2026-08-16
Fixes trailing slash issues for oidc sources that require it.

birdcrazy:
Reported by @jesjhoward in #12
and @RCCore in #5501 

The OIDC issuer URL was getting the trailing slash removed. Some OIDC issuers, like Authentik, require the trailing slash in order to work properly, due to havens exact match requirement.

This PR implements changes which eliminates alteration of the issuer url as entered in the settings.
Also builds out the openid-config path correctly regardless of weather or not the trailing slash is present.

Tested to be working with my Authentik instance.

https://github.com/user-attachments/assets/c8cee26a-fd7f-4f16-9953-ff0413db5f17

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -91,3 +91,6 @@ ADMIN_USERNAME=admin
 # If you choose to leave this feature disabled, Haven will default to Unicode standard 16.0 which is bundled with the server
 # Uncomment to enable this feature
 # UNICODE_EMOJI_AUTO_UPDATE=true
+
+# Your client secret as defined by your OIDC provider. The client secret must be set as an env var if using OIDC.
+#   OIDC_CLIENT_SECRET=changeme-client-secret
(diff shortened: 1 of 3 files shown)


ancsemi:
Merged, thanks for this. Nice catch that the well-known path needs building separately from the issuer itself, and doubly useful that you tested it against a real Authentik instance.

I pushed a follow-up on top of it for two things I noticed while reading it. The comparison now tolerates a missing trailing slash too, so it works whether or not the admin happens to type it the way their provider publishes it. And federated accounts key on the slash-less form, because cfg.issuer is what goes in users.oidc_issuer, so someone who already had SSO accounts and a trailing slash in their settings would have had every returning user land in a brand new account.

Neither is a knock on the PR, they only showed up because your change moved the issuer string around.

PR #5507, opened 2026-08-18
fix(braid): keep main channels out of sub-channel cards

Amnibro:
Reported by **birdycrazy** against both the web client and the Android app: main channels that follow another channel's sub-channels get drawn inside the same Braid card as those sub-channels, so they read as belonging to a tag they have no relation to.

In the report the READ-ONLY card holds `↳ Info and Tips` and `↳ Announcements`, which are sub-channels of **Everyone**, and then `Server Administration` and `Automod Log`, which are main channels. One card reads as one family.

### Cause

`_markRuns()` marks `data-braid-run` from "is my neighbour also a `.channel-item`":

```js
const v = runOf(!prev || !prev.classList.contains('channel-item'),
                !next || !next.classList.contains('channel-item'));
```

Parentage never enters into it, so any unbroken run of rows becomes one card and the first main channel after somebody's subs continues that card.

### Fix

A card is a family: siblings under one parent, or a parent followed by its own subs.

| Neighbours | Same card |
|---|---|
| parent → its own sub | yes |
| sub → sub, same parent | yes |
| sub → sub, different parent | no |
| **sub → main** | **no** (the reported case) |
| main → main | yes |

`famOf()` keys a sub row by `dataset.parentId` and a main row by `main`; `joinsRun()` continues a run only inside one key, or from a parent into its own subs. This leans on two facts from `renderChannelItem` in `app-channels.js`: a sub row always carries `dataset.parentId`, and a parent's subs are always rendered directly after it. The tag row (`sub-channel-item sub-tag-label`) carries the same `parentId`, so a tag stays with the subs it labels.

Cosmetic only — no change to channel data, ordering, collapse state or any other plugin behaviour. Version 1.7 → 1.8.

The Android app had the identical defect in its own drawer and ships the same rule in 7.6.1, with unit coverage over the layout from the report.

### DIFF
--- a/plugins/BraidLayout.plugin.js
+++ b/plugins/BraidLayout.plugin.js
@@ -2,7 +2,7 @@
  * @name Braid Layout
  * @description Vastly simplified two-edge layout: folds the server rail into the sidebar, docks the full voice controls bottom-left, tucks header extras into a kebab menu, merges message runs into cards, and calms the chrome. One-key toggle (Ctrl+Shift+B) between Braid and the classic layout. Suspends itself while Mod Mode edits the layout. Pairs with the Braid / Braid Light themes, and respects every other theme: cosmetic shape rules use :where() so any [data-theme] override wins.
  * @author Amnibro
- * @version 1.7
+ * @version 1.8
  */
 class BraidLayout {
   start() {
@@ -752,10 +752,17 @@ class BraidLayout {
       if (cont) { if (el.getAttribute('data-braid-cont') !== '1') el.setAttribute('data-braid-cont', '1'); }
       else if (el.hasAttribute('data-braid-cont')) el.removeAttribute('data-braid-cont');
     }
+    const famOf = (el) => {
+      if (!el || !el.classList || !el.classList.contains('channel-item')) return null;
+      return el.classList.contains('sub-channel-item') ? `sub:${el.dataset.parentId || ''}` : 'main';
+    };
+    const joinsRun = (a, b) => {
+      const fa = famOf(a), fb = famOf(b);
+      if (!fa || !fb) return false;
+      return fa === fb || (fa === 'main' && fb.slice(0, 4) === 'sub:');
+    };
     document.querySelectorAll('.channel-item').forEach((el) => {
-      const prev = el.previousElementSibling;
-      const next = el.nextElementSibling;
-      const v = runOf(!prev || !prev.classList.contains('channel-item'), !next || !next.classList.contains('channel-item'));
+      const v = runOf(!joinsRun(el.previousElementSibling, el), !joinsRun(el, el.nextElementSibling));
       if (el.getAttribute('data-braid-run') !== v) el.setAttribute('data-braid-run', v);
     });
   }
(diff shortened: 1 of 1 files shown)


Amnibro:
Updated after review feedback: the first rule still let two adjacent **main** channels share a card.

A main channel now **always** starts a new group, so it can never sit inside a run that reads as a sub-channel family. A card is one of: a main channel plus its own subs, a run of sibling subs, or the DM list.

| Neighbours | Same card |
|---|---|
| main → its own sub | yes |
| sub → sub, same parent | yes |
| sub → sub, different parent | no |
| sub → main | no |
| **main → main** | **no** (changed) |
| dm → dm | yes |

`joinsRun()` continues a run only *into* a sub row — from that sub's parent, or from a sibling sub with the same `parentId`. Anything into a main row breaks the run. DM rows (`.channel-item.dm-item`) keep grouping among themselves, since a DM has no sub-channels and the list reads as one block.

Amnibro:
Third commit brings the web in line with the Android app on a second point from the same report.

**A tag header takes over the grouping.** When sub-channels are re-grouped under their own tag label, the label already says what the group is. Indenting the rows underneath it as well hands the sub arrow to channels that only look nested because of where they sit, so the arrow stops meaning "sub-channel".

The core already marks exactly the right rows: `renderChannelItem` sets `dataset.subTag` on a sub only when `subHasTags`, which is the same condition that emits the tag label. So the plugin flags `.channel-item.sub-channel-item[data-sub-tag]` with `data-braid-flat` and paints `#` over the arrow.

- Private subs keep 🔒 — only a literal `↳` is replaced.
- A sub shown directly beneath its parent with no tag header keeps its arrow, which is the case where the indent is the only thing conveying the relationship.
- Cosmetic only, and scoped to `html[data-braid-form="1"]` via CSS, so toggling the form off restores the original glyph with no DOM to unwind.

Android 7.6.2 ships the same rule: children re-grouped under a tag header draw at indent 0 with `#`, children directly under their parent keep indent 1 and `↳`.

ancsemi:
The reported bug is fixed here, a sub followed by a main no longer continues the card, and keying the run by parent is the right idea. Holding off on the merge for two things though.

The main to main case doesn't match your table. `joinsRun` returns early on `fb.slice(0, 4) !== 'sub:'`, so any pair that ends on a main row is false, including main to main:

```
main  -> main    false   (table says yes)
main  -> sub:7   true
sub:7 -> sub:7   true
sub:7 -> sub:9   false
sub:7 -> main    false   (the reported case, correct)
```

So every top level channel ends up `solo` and each one becomes its own pill, which is a fairly big change to how the sidebar reads. If that's actually what you want it's worth saying so in the description, since the table says the opposite. If not, main to main needs to join.

Second, the diff carries a change the description doesn't mention: the `data-braid-flat` block plus the CSS that hides the `↳` and draws a `#` in its place on sub rows. That's a visible change to every sub-channel row, separate from the card grouping fix. Was that meant to be in this PR?

Happy to merge once those are sorted.

ancsemi:
Merged. Cards follow families now instead of whatever happened to be adjacent, which is exactly what birdcrazy reported.

Worth knowing for later: two consecutive main channels no longer share a card either, since only a sub-channel can continue a run. Reads as deliberate to me and it's your plugin, just flagging it in case the intent was for top-level channels to still group.

PR #5508, opened 2026-08-18
Invite Link Fixes

birdcrazy:
There were a few issues I found when testing the invite link system. This PR aims to fix these issues and refine the behavior.

- **Invite links did not open up the registration form automatically.** 
_They instead landed on the login form_
  - Now, when an invite payload is detected in the link, the client switches to the registration form automatically.
- **If server required a registration token, invite links did not bypass the token, and still required the valid token to register.** 
_Since an invite link is a more restrictive and controllable form of granting access than the registration token, an invite link should override the need for the invite token._
  - Now, when an invite payload is detected, the client hides the registration token (as it is now not required for invite links) and sends the invite code to the server side for authentication when the register button is clicked.
  - If the invite code is supplied, the server then checks that the registration code is valid and not fully utilized.
    - Reports appropriate error messages to the user if invalid for any reason.
    - If the invite code is invalid, the client redirects to the login page and removes the invite payload and session parameter; so that the user can then attempt to login as an existing user, or attempt to register with a new invite link, or with the registration token.
  - If the invite code is not supplied (not an invite link), and the admin requires an invite token, the server then checks if the registration token is valid, just like it did in the past.
- **Invite links were, by default, unlimited use, and non-expiring. This could cause invites to be accidentally be created without any restrictions when not intended**
  - Now the defaults have been changed to 1-use, and expiry in 30-days. This way, the link creator must consciously switch these options to unlimited use and/or never-expire when needed.
  
  
Also included is a minor clarification to an error message shown to the user when a username they are trying to register with is already taken. Previously the error message was vague and didn't provide enough information to rectify the error by changing their registration username.

https://github.com/user-attachments/assets/f64515fd-24bf-44e4-92ab-12fdbf78a476

Related to: #5470


birdcrazy:
As a side note, you may want to consider preventing users from registering with (or changing to) passwords that exactly match, or contain their usernames for login security reasons. Since this is what I did in the demo video, I made the password exactly the same as the username..

birdcrazy:
Instead of never allowing invalid invite links to register:

The latest commit alters the registration flow to allow invalid invite links to register when the registration-token isn't required.
If the registration code is required, the invite must be valid in order to register without the registration-token.

ancsemi:
The UX parts here are good. Landing on the login form from an invite link is clearly wrong, and unlimited non-expiring defaults being opt-in rather than the default is a real improvement.

The token bypass needs work before I can merge it though. The check counts rows in `invite_code_uses` to enforce `max_uses`, but registration never writes one. That only happens later, when an already-registered user redeems the invite over the socket (`channels.js`, the `INSERT OR IGNORE INTO invite_code_uses` in the join path). So the count this reads stays at zero through registration, and a 1-use link handed to ten people lets all ten register before anything increments. On a server that turned the registration token on, that's the difference between closed registration and one leaked link being unlimited. Recording the use at registration time, once the user row exists, would close it.

Two other things worth deciding rather than me deciding for you:

`_canManageInvites()` includes anyone holding `invite_users`, so after this change a non-admin can mint links that bypass a gate the code describes as admin controlled. That may well be what you want, since invite links are individually revocable, but it does move where that boundary sits.

The username error going from "Registration could not be completed" to naming that the username exists reverses what looks like a deliberate anti-enumeration message. It's gated behind the token when one is set, but on an open server it lets anyone probe which usernames are taken. Fine if that's a conscious tradeoff for the better error, worth a moment's thought if not.

birdcrazy:
# TLDR:
- Invite usage is recorded when the user enters `/app` and the invite is actually redeemed, rather than during registration.
- New setting `invites_bypass_registration_token` added so the admin can control if invite links are allowed to override the registration token.
- Improved registration flow and error handling.
- Changed "username already exists" error message to be less useful for username enumeration, but still prompts the user to edit their username in the form

# Full comment
## Error Message Refinement

3776d65 Changes the error message a user sees when a they try to register with a username that already exists, The displayed error message is now specific enough to prompts the user to change their username in the form, but vague enough that it doesn't explicitly reveal the usernames that already exist to an attacker probing.  
  
The error message now says “Registration could not be completed: invalid username” instead of the previously explicit “Registration could not be completed: Username already exists” or the original overly vague “Registration could not be completed” message.

However, since this is the only place in the code that this specific error message is shown, an attacker who has access to the code can determine that this error actually means that the username already exists. This was also possible with the original error message.

---

## Registering with an Invite Link

After registering (or an exiting user logging in with an invite link), the invite use does actually get recorded. when the user enters the /app and the channel list is fetched.  
If It was written upon registering, it would potentially utilize the invite usage count before the user is added to the invite channel targets, and/or record more uses than were actually used.  
  
If I understand the flow correctly, the following is my traced path on how the invite gets recorded as used

Upon a successful registration validation, the user is then directed to the app with the invite payload attached to the URL
_public/js/auth.js_

```javascript
const _appQuery = (() => {
    const parts = [];
    if (_pendingInvite) parts.push('invite=' + encodeURIComponent(_pendingInvite));
    if (_pendingChannel) parts.push('channel=' + encodeURIComponent(_pendingChannel));
    if (_pendingMessage) parts.push('message=' + encodeURIComponent(_pendingMessage));
    return parts.length ? '?' + parts.join('&') : '';
  })();
  const _appUrl = '/app' + _appQuery;
  
  ... 
	
  // ── Login ─────────────────────────────────────────────
  loginForm.addEventListener('submit', async (e) => {
    ... // validates registration, returns early if invalid

    window.location.href = _appUrl;
    } catch (err) {
      showError(t('auth.errors.connection_error'));
    }
  });
```

Upon entering the /app, either after registration or existing user login, the user’s channels are fetched:  
_public/js/app.js_

```javascript
class HavenApp {
...
  _init() {
    ...

    this.socket.emit('get-channels');
```

  
_src/socketHandlers/channels.js_

```javascript
module.exports = function register(socket, ctx) {
  ...

  // ── Get user's channels ─────────────────────────────────
  socket.on('get-channels', () => {
    const channels = getEnrichedChannels(
      socket.user.id,
      socket.user.isAdmin,
      (room) => socket.join(room)
    );
    socket.emit('channels-list', channels);
  });
```

  
While fetching the channels, the invite url invite key is fetched (if present) and the user is added to the invite channels via \`socket.emit(‘join-channel‘,..\`, and the invite use is written.

_public/js/modules/app-socket.js_

```javascript
this.socket.on('channels-list', (channels) => {
    ...

    // Auto-join via invite link (vanity code or channel code in query param)
    const urlParams = new URLSearchParams(window.location.search);
    const inviteCode = urlParams.get('invite');
    if (inviteCode && !this._inviteHandled) {
      this._inviteHandled = true;
      this.socket.emit('join-channel', { code: inviteCode });
      sessionStorage.removeItem('haven_pending_invite');
      // Clean up the URL
      const cleanUrl = window.location.pathname;
      window.history.replaceState({}, '', cleanUrl);
    }
```

  
And then here is where the invite use is recorded.  
_src/socketHandlers/channels.js_ 

```javascript
  // ── Join channel by code ────────────────────────────────
  socket.on('join-channel', (data) => {
    ...

    // ── Managed invite link (multi-code menu) ──────────
    // Checked before the legacy vanity/server codes and before the hex-format
    // gate, since a managed code can be a custom slug. Each carries its own
    // channel grant, on/off switch, and optional expiry by time / distinct-user
    // count. A user who already redeemed a code can re-enter it freely (e.g. to
    // refresh) without burning another use against the limit.
    const inviteRow = db.prepare(
      "SELECT *, (expires_at IS NOT NULL AND expires_at <= CURRENT_TIMESTAMP) AS is_expired FROM invite_codes WHERE code = ?"
    ).get(code);

		...

    if (inviteRow) {
			db.prepare(
        'INSERT OR IGNORE INTO invite_code_uses (invite_code_id, user_id) VALUES (?, ?)'
      ).run(inviteRow.id, socket.user.id);
      socket.emit('channels-list', getEnrichedChannels(socket.user.id, socket.user.isAdmin, (room) => socket.join(room)));
      socket.emit('error-msg', `Invite accepted — joined ${joinedCount} channel${joinedCount !== 1 ? 's' : ''}`);
      return;
    }
			
```

## Invite Link Overriding Registration Token

The following is assuming the scenario that the registration token is required by the admin.  
  
If the registration token requirement is not bypassed for invite links when the registration token is required, then new users will need both the invite link, and the registration token, Making the invite link less useful when a non-admin invites users, as the admin will still need to share the registration token, or the inviter needs to know and share the token; An inconvenient and unnecessary extra step in my opinion.  
  
5d2e0d6 adds a new setting so the admin can decide if they want to allow invite links to override the registration-token. The setting is used to appropriately determine the visibility of the registration-token field; as well as appropriately considering the settings value in the backend registration flow.  
The redirect is now removed, as that would require the user to re-fill the form (an annoyance to avoid if possible). Instead, when the register button is pressed on an invalid invite link, the invite code is stripped from the url and memory, and the registration token field is displayed if required. This allows the user to retain the filled out form, and attempt to register again with the token.

birdcrazy:
# Summary
- Changed "username already exists" error message to be less useful for username enumeration, but still prompts the user to edit their username in the form
- Invite usage is now written when the user successfully registers.
- Added a new setting `invites_bypass_registration_token`. this allows the admin to control if invite links are allowed to override the registration token.
- Improved registration flow and error handling.

# video demo:
https://github.com/user-attachments/assets/36fc9ccb-b432-4458-997d-a8695086c473

# Full Explanation
## Error Message Refinement

3776d65 Changes the error message a user sees when a they try to register with a username that already exists, The displayed error message is now specific enough to prompts the user to change their username in the form, but vague enough that it doesn't explicitly reveal the usernames that already exist to an attacker probing.  
  
The error message now says “Registration could not be completed: invalid username” instead of the previously explicit “Registration could not be completed: Username already exists” or the original overly vague “Registration could not be completed” message.

However, since this is the only place in the code that this specific error message is shown, an attacker who has access to the code can determine that this error actually means that the username already exists. This was also possible with the original error message.

---

## Registering with an Invite Link

3355857: once registration is successful and the user is written, the invitation usage then also gets written during the registration flow.

## Invite Link Overriding Registration Token
  
If the registration token requirement is not bypassed for invite links when the registration token is required, then new users will need both the invite link, and the registration token, Making the invite link less useful when a non-admin invites users, as the admin will still need to share the registration token, or the inviter needs to know and share the token; An inconvenient and unnecessary extra step in my opinion.  

But I agree that the admin should be aware that invite links could bypass the registration code. An admin who doesn't know or want this, could experience people joining their server that they do not approve of.
  
5d2e0d6 adds a new setting so the admin can decide if they want to allow invite links to override the registration-token. The setting is used to appropriately determine the visibility of the registration-token field; as well as appropriately considering the settings value in the backend registration flow.

The redirect is now removed, as that would require the user to re-fill the form (an annoyance to avoid if possible). Instead, when the register button is pressed on an invalid invite link, the invite code is stripped from the url and memory, and the registration token field is displayed if required. This allows the user to retain the filled out form, and attempt to register again with the token.

ancsemi:
Merged, thanks! The part I looked hardest at was whether registering with an invite would burn the use twice, once at signup and again when the app redeems the same code on load. It doesn't. The redeem path checks alreadyUsed before the max_uses gate, so a one-use link doesn't lock out the very person who just used it, and the insert is OR IGNORE against the composite key so the count stays at one.

Nice touch putting the username check after the captcha and invite validation, and having the client drop a bad invite from the URL so people can fall back to the token instead of being stuck.

Heads up that the new link form defaults are now 1 use and 30 days instead of unlimited and never. I'm happy with that as a default, just noting it since anyone used to the old form will see different numbers.

PR #5510, opened 2026-08-18
[feat] Border / PFP Overlay + [QOL] No more animated avatar spam

Bo0sted:
## [feat] Border / PFP Overlay + [QOL] No more animated avatar spam

Two related additions to profile pictures: decorative frame overlays with a
fitting editor, and a per-user policy for how animated avatars play. Animated
avatars no longer loop constantly without condition. Added focus condition
to match how Discord handles animated profiles. 

### Border / PFP overlays

Upload a decorative frame and fit it over your avatar in an event-sourced
Edit modal. The saved fit then renders on every avatar everywhere: chat
messages, the user list, and the profile popup. The fit travels with each
message, so it shows for offline authors too.

**Editor tools:** Crop, Move, Resize, Rotate, Distort, Opacity. Each edit is
appended to an op log stored as size-independent fractions, so one saved fit
renders correctly at every avatar size. Each edit composes on top of the ones
already applied, so they stack predictably.

- Distort is guarded so no perspective can fling content across chat; invalid
  quads render flat instead of saving a broken homography.
- "Crop out invisible pixels" auto-fits the crop to the image's opaque bounds,
  measured per side so uneven frames crop correctly.
- Rotate has +/-90 buttons alongside free rotation.
- Resize anchor (Center/Corner): Center resize makes the border grow
proportionally in all directions, corner mode locates which corner is nearest
to the edge of the avatar, and anchors it, so a resize grows the opposite direction
(into the avatar frame) instead of drifting outward
- A resize that would drift the frame clearly out of frame is refused with a
  toast instead of persisting a broken fit.
- The edit/commit log on the side of the Edit modal has an undo button, and clicking
undo on a commit will undo every commit up until that point including the one
you're undoing. 

**Persistence and safety:** new `border` and `border_transform` columns, a
`POST /api/set-border-transform` endpoint, and a socket broadcast. The op log
is sanitized server-side (strict allowlist, every field coerced to a clamped
finite number, unknown ops dropped, log length capped) on both write and read,
then folded into nested CSS wrappers at each render site. All CSS is built from
numeric values only, so nothing arbitrary reaches a style attribute.

### No more animated avatar spam (per-profile animation policy)

An "Animated profile" control in Edit Profile lets you choose how your animated
avatar and border play for everyone:

- **Trigger:** plays on message hover or when the profile card is open.
- **Disabled:** shows the first frame only.

Freezing is pure client side: the first frame is captured to a canvas data URL
and swapped in through a single body-level observer, so no static file or
server image work is needed. Same-origin uploads keep the canvas untainted; if
a privacy mode blocks the read, the image is left animated. The policy is
persisted via a new `animate_profile` column and `POST /api/set-animate-profile`,
and threaded through sessions, messages, and profile data so offline authors'
avatars honor it too.

### Other

- Border-editor handle tooltips and the border upload toasts are localized;
  new keys added to `en.json` with graceful English fallback for locales that
  do not have them yet.
- Idempotent migrations add the `border`, `border_transform`, and
  `animate_profile` columns.

### Testing

- Registered a fresh admin, uploaded an avatar and a frame, exercised each
  editor tool, and confirmed the fit renders on messages, the user list, and
  the profile popup.
- Tried to save an invalid border by moving it to a corner, resizing, distorting and moving.
- Verified the animation policy freezes and plays per Trigger/Disabled.
- Confirmed the new i18n keys resolve and the app loads with no console errors.


ancsemi:
This needs a rebase before it can go anywhere. Main moved today (a channel thread list, Unicode display names, an upload disk guard) and it collides across app-ui.js, app-socket.js, en.json, and three files under src/socketHandlers.

On the code itself, I read the sanitizer, the CSS builder, the upload endpoint and every emit site closely, and the security core is solid. Strict op allowlist, every field coerced to a clamped finite number, unknown ops dropped, the log capped, and sanitized on read as well as write rather than trusting what's in the database. `_borderWrapperStyle` only ever interpolates numbers into the style attribute, so there's no way through it. Mirroring the avatar upload for the magic byte checks was the right instinct too.

Two things to pick up when you rebase:

`/api/upload-border` doesn't check the `upload_files` permission, while `/api/upload` does. Worth matching so someone who has had uploading taken away can't still push files through the border route.

There's a new `uploadDiskGuard` middleware on main that sits between `uploadLimiter` and the handler on every upload route. The border route wants it too, otherwise it's the one path that can eat into the reserved disk space.

I haven't gone line by line through all 900-odd lines of the editor yet. I'll do that once it's rebased, since a lot of it will move anyway.

Bo0sted:
Done. Looking forward to your feedback.

ancsemi:
Merged, thanks Matt. The border editor is a lot of careful work and the op log approach (fractions, sanitized server side, folded into nested wrappers) is a nice way to make one saved fit render right at every avatar size.

One thing I added on top after merging: the animation policy only had your side of it, where the person with the animated pfp decides. Someone who just doesn't want to see moving avatars had no say. So there's now a viewer side too, in Settings under Layout, with Always, On hover, and Never. The stricter of the two wins, so anyone who sets their own profile to Disabled still stays frozen for everyone regardless of what a viewer picks, and Always gives back the constant looping for people who liked it that way.

Also worth a look when you get a chance: `_animCanAnimate` matches `.png`, so every ordinary static PNG avatar gets fetched again, drawn to a canvas and swapped for a data URL even though there is nothing to freeze. It's deduped per URL so it isn't per message, but it's still a fair bit of work for no result on the most common avatar format. Sniffing for an `acTL` chunk, or just dropping png from that regex, would skip it.

PR #5511, opened 2026-08-18
Fix role editors silently doing nothing on servers that predate an event

Amnibro:
## What happened (community report)

A self-hoster reported that editing the **Admin role's name/color did nothing** — no toast, no console error, nothing — while editing every other role saved fine. After ~45 minutes of debugging it turned out their server.js predated 3.44.0, while their client files were current: `update-admin-role-display` (added in 3.44.0) simply has **no handler** on their server, and socket.io never delivers an ack for an unhandled event, so the Save callback never ran. Normal roles kept working because `update-role` is much older.

This mismatch is easy to hit: the server serves `public/` straight from disk, so anyone who updates files without fully restarting/updating server.js runs new client code against an old server — and every role-editor button that depends on a newer event turns into a dead button.

## Fix

All acked role-editor emits now go through a `_roleEmit` wrapper built on `socket.timeout(10000)`. A missing handler surfaces after 10s as a clear toast:

> No response from the server — it may be running an older Haven version. Update the server files and fully restart it, then try again.

New `toasts.role_server_no_response` string added to all 7 locales (`validate-locales` passes).

## Verified live

- **Pre-3.44 server.js + these client files** (the reporter's exact setup, reproduced with a worktree at `5d3a85d~1`): unfixed client = silent no-op on Save, fixed client = toast at 10s — including immediately on opening the Admin role editor, which tells the owner right away their server is behind.
- **Current server + these client files**: Save still works normally (green "Role saved" toast in ~1.3s, name persists).


ancsemi:
Merged, thanks!

ancsemi:
This is in. The timeout wrapper is a good shape for it, and I checked that revoke-role still acks on the one call site that passes no callback, so a current server will not trip the toast by accident.

One small thing changed on top: the new string had an em dash in all seven locales and house style has none in user-facing text, so each one is two sentences now. Nothing wrong with the translations themselves.

PR #5512, opened 2026-08-18
Add view_all_channels role permission — mods can auto-see every channel

Amnibro:
## Community ask

From a self-hoster running a community server:

> if there is a "Server Wide" role system, such as for Mods or Admins, they should automatically see ALL channels, i have a few mods in my spot and they can't see anything they gotta add it manually

Role-channel-access linking covers the channels you remember to tick, but there was no way to say **"this role simply sees everything — now and in the future"**, which is how Discord-style Mod/Admin roles behave and what server owners expect. Every new channel meant another manual round of adding each mod.

## What this adds

A new role permission, **`view_all_channels`** ("See All Channels (Auto-Join)"):

- Holders get the admin's visibility treatment in `getEnrichedChannels`: every non-DM channel, membership filled in on the fly — so **channels created after the grant appear automatically**, zero bookkeeping.
- Assigning a role that carries it (or editing a role to add it) **pushes a rebuilt channel list to the holder's live sockets immediately**, the same mechanism linked channel access already uses.
- Because it reveals private channels, it's **admin-only to grant** — added to `adminOnlyPerms` on the server (all three enforcement sites) and `ADMIN_ONLY_PERMS` in the client, alongside `transfer_admin`/`manage_roles`/`manage_server`/`delete_channel`.
- Label added to all 7 locales (`validate-locales` passes).

Revoking the role stops future auto-adds; memberships already created stay (same semantics as admin — prune via the existing member management if needed).

## Verified live on a scratch server

- Non-admin with the role: private channel appeared **instantly** on assign (live socket push, no reload), could open and post in it.
- A private channel created **after** the grant showed up for them automatically.
- A user without the role saw neither private channel.
- Existing admin behavior unchanged; `node --check` passes on all touched server files.


ancsemi:
Merged, thanks. This is the right answer to the ask and the admin-only gate on granting it is the correct call, since it exposes private channels.

I pushed a follow-up for the revocation side. Because the permission writes a real channel_members row per channel, losing it needed to take those rows back, and it didn't: revoking the role, editing the permission off it, or deleting the role all left the holder in every private channel. Demoting a mod took their badge and nothing else.

Rows the permission creates are now marked with a column on channel_members so the cleanup can tell them from memberships someone got another way, which stay put. It runs on revoke, on an edit that drops the permission, and on delete, and no-ops if they still hold it via another role. I also made the sockets leave the channel rooms, since otherwise they keep receiving live messages from those channels until they reconnect.

Admin auto-joins are deliberately left unmarked, since admin isn't revoked the same casual way a mod role is.

PR #5513, opened 2026-08-18
Refuse to boot when a stale pre-split module file shadows a module folder

Amnibro:
## The crash (community report, v3.47.0)

```
/home/raidenphantom/Haven/server.js:4427
}).activity;
   ^
TypeError: Cannot read properties of undefined (reading 'activity')
```

Same self-hoster as the role-editor thread. His server.js **is** current — line 4427 is `activityRef.engine = setupSocketHandlers(...).activity`. The crash means `setupSocketHandlers()` returned `undefined`… which the current module can't do. It returns `{ activity }`.

## Root cause: a deleted file from 2.9.7 still on disk

`src/socketHandlers.js` was split into the `src/socketHandlers/` folder in **2.9.8**. His update style is copying new files over the existing install — which never deletes anything — so the pre-split monolith is still there. Node resolves `require('./src/socketHandlers')` to the **file** before the **directory**, the monolith exports the same `{ setupSocketHandlers, sanitizeText }` shape, its `setupSocketHandlers` registers all the ancient handlers and **returns nothing** → `undefined.activity`.

The nasty part is what happened *before* the crash: with his old server.js (which never read `.activity`), this booted fine — meaning his server's **entire socket layer has been frozen at ≤2.9.7 for months** while every file around it was current. That one stale file retroactively explains his whole support history: the Admin-role Save doing nothing (`update-admin-role-display`, added 3.44.0, never loaded), embed reactions behaving differently than the public Haven server, the desktop app finding API paths missing. Any self-hoster who installed ≤2.9.7 and updates by unzip-over-the-top has this same landmine.

## Fix

A startup guard right after the Node version check: scan `src/` for any `<name>.js` file that shadows a `<name>/index.js` folder, and exit with the exact file list and instructions (delete the file / update by replacing the folder; data in the data dir is safe) instead of booting mixed-version code. Generic, so a future file→folder split is covered automatically.

## Verified

- Dropped the 2.9.7 `socketHandlers.js` (`git show fad2a5d~1:src/socketHandlers.js`) into a current tree: **reproduces the reported crash byte-for-byte** (same line 4427, same TypeError).
- With the guard: same stale tree now prints the file name + fix instructions and exits 1.
- Clean tree: boots normally, `/api/health` OK. `node --check server.js` passes.

## For the reporter

Immediate unblock, no new release needed: `rm /home/raidenphantom/Haven/src/socketHandlers.js` (on both of his installs), then start the current server.js. Worth re-checking the embed-reaction behavior after that — his server will be running its first current socket layer since April.

### DIFF
--- a/server.js
+++ b/server.js
@@ -14,6 +14,41 @@ if (nodeMajor < 22 || nodeMajor > 26) {
 // Bootstrap .env into the data directory if it doesn't exist yet
 const fs = require('fs');
 const path = require('path');
+
+// ── Stale-install guard ───────────────────────────────────
+// Updating by unzipping/copying a release over an existing install leaves
+// behind files that newer versions deleted. That is normally harmless — until
+// the deleted file is a module that was split into a folder of the same name
+// (src/socketHandlers.js became src/socketHandlers/ in 2.9.8): require()
+// resolves the leftover FILE before the directory, so the server silently
+// runs months-old module code no matter how current every other file is, and
+// eventually dies somewhere unrelated. A real self-host crashed on boot with
+// "Cannot read properties of undefined (reading 'activity')" because a
+// pre-2.9.8 socketHandlers.js was still shadowing the folder — after months
+// of its socket layer being frozen at the old version while "fully updated".
+// Catch the pattern generically and say exactly which file to delete.
+{
+  const srcDir = path.join(__dirname, 'src');
+  let entries = [];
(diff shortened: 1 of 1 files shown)


ancsemi:
Merged, thanks!

ancsemi:
This is in. Good bit of detective work, and the generic form is the right call, since the next file to folder split gets covered for free rather than needing another special case.

The part worth flagging for anyone reading later: the reason this was so hard to spot is that nothing looked wrong. Every file was current, the version string was current, and the socket layer had been frozen for months underneath it.

PR #5514, opened 2026-08-19
[QOL] Remove the Google connections users can't opt out of

Bo0sted:
### One of Haven's design philosophies is no Big Tech, and this commit aims to further align Haven to truly being independent of Big Tech.

Two things reached out to Google on their own, with no way for an admin or member to stop them:

- Fonts loaded from fonts.googleapis.com / fonts.gstatic.com on every page load. They are now bundled in public/fonts and served locally with a long immutable cache, so each client downloads them once. The Google Fonts hosts are removed from the CSP, which now blocks them outright.
- Google's public STUN server (stun.l.google.com) sat in the default voice ICE pool. It is dropped from both the server default and the client fallback. Cloudflare, Metered and Twilio remain (all free, no credentials, verified working), and voice still degrades gracefully if they ever fail.

This commit intentionally leaves the Google connections a user CAN control, and makes that choice visible rather than removing it:

- FCM mobile push: a new Settings > Security > FCM Privacy toggle. Default on so nothing breaks, but an admin can turn it off, and the note points to UnifiedPush as a Google-free alternative.
- YouTube embeds and previews: already disableable through Automod domain rules.
- Tenor GIF search: already optional, and Giphy can be used instead.

No functional change for anyone who leaves the defaults.


ancsemi:
Merged. Before pulling the font files in I hashed all four against what Google Fonts serves right now and they match byte for byte, and the unicode-range lines match too, so glyph coverage is unchanged. Tightening the CSP in the same pass was a good call since it means a Google font URL can't quietly creep back in later without someone noticing.

I booted a throwaway instance to check it end to end: fonts serve from /fonts with the year long cache, style-src and font-src come back as 'self' only, and flipping the FCM toggle off does show the "turned off in Settings" line on boot instead of the enabled one.

One thing I can't check from this side is the claim that Haven Mobile prioritises UnifiedPush over FCM. Taking your word on it since that's the app side, just flagging that if it ever stops being true the hint text needs a reword.

Nice work on this one.

PR #5515, opened 2026-08-19
[fix] Last.FM album art incorrect subdomain

Bo0sted:
Last.fm serves album art from the lastfm-img subdomain, but the image host allowlist only had lastfm.freetls.fastly.net, so safeImage rejected every current artwork URL and the listening card showed no image.


ancsemi:
Merged. Last.fm started serving art off that subdomain and the allowlist never picked it up, so thanks for tracking down why the covers were coming back blank.

PR #5517, opened 2026-08-21
Add bot voice and audio APIs

bernardokcosta:
## Summary

- add dynamic MP3, WAV, and OGG playback with per-bot queues, skip, stop, capability URLs, and temporary-file cleanup
- add a Socket.IO/WebRTC voice gateway with visible bot presence and signaling support
- add voice-channel presence and bulk message deletion endpoints
- add the admin-only `can_use_voice` webhook permission
- route soundboard playback through voice rooms and respect deafen state
- normalize callback metadata and provide backward-compatible HMAC signature headers
- migrate runtime, persistent, and client state when channel codes rotate
- document the expanded bot API and voice gateway

## Security

- revalidate webhook permissions after uploads and during active sessions
- isolate bot sockets from text rooms and namespace-wide administrative broadcasts
- prevent non-admins from moving privileged webhooks between channels
- pin callback requests to DNS-validated public addresses to prevent SSRF and DNS rebinding
- enforce callback deadlines across DNS resolution and HTTP delivery
- restrict audio downloads to short-lived playback capabilities
- clean queued audio when permissions are revoked or channels are deleted
- exclude temporary bot audio from backups

## Dependencies

- add `music-metadata` for audio validation and duration inspection
- upgrade Multer to 2.2.0
- align the declared runtime requirement with Node.js 22+

## Testing

- `npm test` with 12 passing tests
- integration coverage for REST endpoints, upload revocation, and aborted uploads
- unit coverage for audio queues, global queue limits, file cleanup, signaling, access revocation, duplicate bot connections, full channels, and channel rotation
- `node --check` passed for modified backend and test files
- `git diff --check` passed
- final security review found no remaining high or medium blockers


ancsemi:
Read through all of this. The engineering is solid and this isn't a no, but there are three things I'd need changed before it can go in, and one piece I think deserves to land on its own much sooner than the rest.

**The engines bump.** This raises the floor to node >=22, and I don't think anything in the PR actually needs it. music-metadata 11.15.0 declares >=18 itself, and you're loading it through a dynamic `import()` rather than `require()`, which is exactly the pattern that works for an ESM package on Node 18. So as written this drops Node 18 and 20 self-hosters for no gain. Can that go back to >=18.0.0?

**The soundboard room change.** The bot sound endpoint moves from `channel:${code}` to `voice:${code}`. The catch is the client handler also gates on `data.channelCode === this.currentChannel`, so after the change a listener has to be in the voice room *and* looking at that channel to hear anything. Right now they only have to be looking at the channel. Anyone with a soundboard bot today will experience that as it breaking. If the goal is to make soundboard a voice thing, that's a conversation worth having on its own, but it shouldn't ride along inside a voice API PR.

**The size.** It's around 2600 lines over 28 files, and some of it isn't bot voice. The channel code rotation extraction into src/channelRotation.js is a separate concern, and package.json comes back without its BOM, which our release process puts back anyway. Splitting the bot voice and audio APIs from the rest would make this reviewable, and honestly it'd make it much easier to say yes to.

The piece I'd like to see sooner: the webhook callback hardening. Pinning the validated address through a custom `lookup` so a second DNS answer can't rebind the request is a real fix, and the current `isSafeCallbackUrl` genuinely does miss bracketed IPv6 loopback, 127.x beyond 127.0.0.1, and credentials in the URL. That's a security fix sitting behind a large feature review, which isn't where it should be. If you want to pull just that out into its own PR I'll look at it right away.

Thanks for the tests too, those made reading the rest a lot easier.

bernardokcosta:
> Read through all of this. The engineering is solid and this isn't a no, but there are three things I'd need changed before it can go in, and one piece I think deserves to land on its own much sooner than the rest.
> 
> **The engines bump.** This raises the floor to node >=22, and I don't think anything in the PR actually needs it. music-metadata 11.15.0 declares >=18 itself, and you're loading it through a dynamic `import()` rather than `require()`, which is exactly the pattern that works for an ESM package on Node 18. So as written this drops Node 18 and 20 self-hosters for no gain. Can that go back to >=18.0.0?
> 
> **The soundboard room change.** The bot sound endpoint moves from `channel:${code}` to `voice:${code}`. The catch is the client handler also gates on `data.channelCode === this.currentChannel`, so after the change a listener has to be in the voice room _and_ looking at that channel to hear anything. Right now they only have to be looking at the channel. Anyone with a soundboard bot today will experience that as it breaking. If the goal is to make soundboard a voice thing, that's a conversation worth having on its own, but it shouldn't ride along inside a voice API PR.
> 
> **The size.** It's around 2600 lines over 28 files, and some of it isn't bot voice. The channel code rotation extraction into src/channelRotation.js is a separate concern, and package.json comes back without its BOM, which our release process puts back anyway. Splitting the bot voice and audio APIs from the rest would make this reviewable, and honestly it'd make it much easier to say yes to.
> 
> The piece I'd like to see sooner: the webhook callback hardening. Pinning the validated address through a custom `lookup` so a second DNS answer can't rebind the request is a real fix, and the current `isSafeCallbackUrl` genuinely does miss bracketed IPv6 loopback, 127.x beyond 127.0.0.1, and credentials in the URL. That's a security fix sitting behind a large feature review, which isn't where it should be. If you want to pull just that out into its own PR I'll look at it right away.
> 
> Thanks for the tests too, those made reading the rest a lot easier.

Thanks for the detailed review. I agree with all four points.

I'll restore the original Node engine range, remove the soundboard behavior change, and split the unrelated work into independent PRs.

I'll start by extracting only the webhook callback SSRF/DNS-rebinding hardening into a small PR against the latest upstream main. After that, I'll separate channel rotation, the voice gateway, dynamic audio, and bulk deletion so each can be reviewed independently.

I'll preserve the existing package.json BOM and keep callback payload/signature behavior unchanged in the security PR.

PR #5520, opened 2026-08-22
Harden webhook callback delivery against SSRF

bernardokcosta:
## Summary

Hardens webhook callback delivery against SSRF and DNS rebinding while preserving the existing callback API and `HAVEN_ALLOW_PRIVATE_CALLBACKS` behavior.

## Changes

- reject callback URLs with embedded credentials or unsupported protocols
- validate IPv4, IPv6, IPv4-mapped, transition, reserved, and non-global address ranges
- inspect every address returned by DNS and reject mixed public/private results
- pin the HTTP connection to the exact address that passed validation
- preserve the original hostname for the Host header and TLS SNI
- apply one deadline across DNS resolution and HTTP delivery
- avoid following redirects by using the Node HTTP/HTTPS transports directly
- keep link-local and metadata destinations blocked when private callbacks are enabled
- avoid retrying destinations rejected by the SSRF guard

## Compatibility

- preserves callback payloads
- preserves existing HMAC signature headers
- preserves retry behavior for network errors and 5xx responses
- preserves `HAVEN_ALLOW_PRIVATE_CALLBACKS` for loopback, RFC1918, and ULA destinations
- does not change `package.json`, the Node.js engine range, or dependencies

## Testing

- URL credentials and unsupported protocols
- decimal and hexadecimal loopback representations
- IPv4 and IPv6 private, loopback, link-local, reserved, and transition ranges
- IPv4-mapped and IPv4-compatible IPv6 forms
- mixed public/private DNS responses
- private callback opt-in behavior
- DNS address pinning without a second resolution
- UTF-8 payload length and exact request body
- DNS resolution deadline

## Verification

- `node --test test/webhookCallback.test.js`: 6 passing
- tests pass on Node.js 18.0.0
- `node --check` passed
- `git diff --check` passed

### DIFF
--- a/src/socketHandlers/index.js
+++ b/src/socketHandlers/index.js
@@ -15,6 +15,11 @@ const { socketClientIp } = require('../clientIp');
 const automod = require('../automod');
 const { resolveSpotifyToYouTube, searchYouTube, fetchYouTubePlaylist, extractYouTubeVideoId, resolveMusicMetadata } = require('./musicResolver');
 const createPermissions = require('./permissions');
+const {
+  UnsafeCallbackError,
+  postWebhookCallback,
+  validateCallbackUrl
+} = require('../webhookCallback');
 
 const { createActivity } = require('../activity');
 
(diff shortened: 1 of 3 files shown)


ancsemi:
Merged, and this is exactly the split I was hoping for. Thanks for turning it around so fast.

I ran your tests plus about 40 cases of my own against it. The ones I most wanted to see fail did: decimal, hex and octal loopback encodings, `::ffff:127.0.0.1` and the raw `::ffff:7f00:1` form, credentials in the URL, and a rebind where a public hostname answers with 127.0.0.1. Also checked that a DNS answer mixing a public and a private address gets rejected in either order, not just when the private one comes first.

The case I was most concerned about was someone on a cloud host flipping HAVEN_ALLOW_PRIVATE_CALLBACKS on and accidentally opening their metadata endpoint, since `.internal` passes the syntax check when the opt-in is enabled. It resolves to 169.254.169.254 and gets caught at the DNS stage, so that hole is closed. Good call keeping link-local out of the opt-in entirely.

Making IPv6 default-deny outside 2000::/3 rather than blocklisting known-bad ranges was the right way round, and dropping the retry for URLs the guard rejected saves a pointless second attempt.

100.64/10 in the opt-in list was a nice thought for anyone running their bot over Tailscale.

Issue #5444, opened 2026-07-20
Non desktop users get frequently kicked from voice chat


AlexT2803:
Not my first time opening this but im trying to finally get it resolved as people are resorting to using steam vc.

As it currently stands: People who are on the web client get disconnected at random time invervals ranging from less than 30 seconds to a few minutes. SInce the last fix other people in call can now reliably hear the person who got kicked after they rejoin but now it seems like some people are reporting they cant hear other people in the call on rejoin and have to manually rejoin. 

My current setup is haven and coturn and this issue has persisted after migrating my entire (perviously very janky) setup to a clean vps.

ancsemi:
Pushed a change aimed at the "can't hear other people after rejoining" half of this. Before, when you dropped and came back, the app made a single attempt to repair the audio link to each person, and if that attempt didn't take it just gave up and left them silent until you fully left and rejoined. Now it double checks a few seconds later and keeps retrying the repair, so fewer of those cases should need a manual rejoin.

What I still want to figure out is why the web client keeps dropping every 30 seconds to a couple minutes in the first place, since it shouldn't be doing that at all. Two things that would help narrow it down: is Haven sitting behind something like nginx, Caddy, or a Cloudflare tunnel? Those tend to close idle websockets on a short timer, and it would only hit browser users, not the desktop app. And when someone gets dropped, does the browser console (F12) actually show the socket disconnecting, or does the call just go silent while everything still looks connected?

I don't want to call it fixed until you can confirm it actually helps on your end, so let me know how it behaves once you're on this update.

AlexT2803:
I am running haven behind Caddy. Ill try to get some people on to test with the browser console open but atleast on havens end the client is disconnecting because it gives everyone in call the leave and join sounds and double notifications.

AlexT2803:
<img width="1270" height="170" alt="Image" src="https://github.com/user-attachments/assets/0ee9069b-f460-453c-b4ec-ce7ab7c1ff4a" />

ancsemi:
Appreciate the details. Two separate things here.

[...]

Issue #5427, opened 2026-06-23
Web client users are kicked from voice chat and instantly reconnect with no audio


AlexT2803:
Haven version: v3.26.0
Turn server: selfhosted coturn

People will be talking in chat when suddenly one person will appear to leave and rejoin almost instantly. After this happens some people may or may not be able to hear the affected person. In some instances certain people will be able to hear and be heard by the affected person while others in the same call will not be able to hear or be heard by the affected person. To fix the issue the person who disconnects usually has to disconnect and reconnect at which point everything works perfectly. The client that is getting disconnected doesn't have any indication that anything has happened. 

As far as my server can tell this is only happening to people who use the web client with both Firefox and Edge users being affected. For one person they had no issues using the desktop client but encountered issues when they got a new pc and connected using the web client. Even for two people on the same network the one on the desktop client will not have the issue while the one on the web client does. The disconnects dont appear to be at regular intervals but usually happen every 5-10 minutes.

ancsemi:
Is this a new issue as of 3.26.0? 
And, are your channel codes set to static, or rotate dynamically?

averyfaulk:
Happening for us as well on 3.27.0, channel codes are set to static.

ancsemi:
Thanks, the web-only detail is really helpful. I added some recovery logic for the no-audio part: when a web client's connection blips and comes back, the existing peer connections can survive in a half-dead state where some people can still hear you and others can't, which matches exactly what you're describing. After a reconnect it now checks each peer and rebuilds only the broken ones, so audio should come back on its own instead of needing a full disconnect/reconnect.

I want to be upfront that this targets the audio loss, not whatever is dropping the web socket every 5-10 minutes in the first place, I haven't pinned that down yet and it looks environmental (browser or something between the client and server). So once this is in the next build, could you and @averyfaulk test whether the audio recovers on its own now even if the brief disconnect still happens? That'll tell me whether the recovery is working while I keep looking at the root cause.

AlexT2803:
Just pulled latest and it looks like recovery isnt working

AlexT2803:
So when they reconnect voice activity is shown server side but still cant be heard by some people.

ancsemi:
Thanks for testing and reporting back so quickly, that detail ("voice activity shows server-side but some people still can't hear them") is exactly what I needed.

[...]

Issue #5391, opened 2026-05-27
No Channels Visible


birdcrazy:
_(On Web)_
Sometimes when I switch to another server. None of the channels appear.
I assume this is due to the auth token expiring, but it doesn't kick me back to the login screen until manually clicking the logout button.
Sometimes refreshing the page a bunch might eventually send me back to the login screen.

<img width="329" height="781" alt="Image" src="https://github.com/user-attachments/assets/45cd9d4e-4fb3-4cd5-834c-ccd47a9beb5e" />

ancsemi:
Yes, 've been having this issue since the temporary auth implementation as well. I took a crack at it the other day and it _should_ be sending you back to login instead, but obviously something's still off. I'll give it another look.

birdcrazy:
Oh ok. It happened to me a few times in the past week or so. 
This most recent occurrence today was on the community server, which is still on 3.16.13. 

So if your fix was after 3.16.13, the community server just probably doesn't have it yet.

ancsemi:
Took another pass at this. The previous attempt only handled the case where the socket's auth middleware rejected with one of three exact error strings; turns out there are other ways for the session to end up in a "connected but channels never arrive" zombie state (DB hiccup mid-handshake, an exception in the channel enrichment query, user row out of sync, etc.), and none of those fire connect_error.

Added a watchdog on the client: if the socket connects but channels-list doesn't arrive within 10 seconds, it hits /api/auth/validate over HTTP. 401 → kick to login. 200 → retry get-channels once, then force a reload if that also goes nowhere. That should cover whatever path was leaving you stranded before.

Will be in the next release.

ancsemi:
Whoops, didn't mean to close this one yet! 

Leaving open until we can confirm it's fixed.

birdcrazy:
I have not had this occur again on my servers. It seems to kick me back to the login page correctly.
Feel free to close this if you are also satisfied with the current behavior.

ancsemi:
Thanks for confirming!

Issue #5347, opened 2026-05-07
Users in call cant see anyone in calll


AlexT2803:
When you are in a call you can no longer see other people (or yourself) in the call unless they joined before you. Im in a call with 2 other people here.

<img width="239" height="367" alt="Image" src="https://github.com/user-attachments/assets/be0e129f-23ed-4250-859d-a4921281305a" />

RiffyDivine:
We are seeing the same issue and also those who aren't showing up in the room can't be heard in voip by others but they can join the call.

ancsemi:
Dug into this and found three separate bugs combining to produce exactly what you're describing. Pushed a fix in 3.14.1.

What was happening:

1. The server's voice-rejoin handler (which fires when your socket reconnects after a network blip) was silently overwriting your slot in the voice room without telling the rest of the room. Everyone else still had a dead `RTCPeerConnection` to your old socket session, and when your client sent a fresh offer it got applied on top of the dead ICE - so audio never recovered. Now it cleans up the old session properly, broadcasts the leave, and re-adds you so the room rebuilds with a clean handshake.

2. Even if the server had cleaned up correctly, the client was accepting the rebuilt offer onto a peer connection whose `connectionState` was already `failed` / `closed`. It now tears down and rebuilds the peer when it sees that.

3. The 2-second retry on `voice-leave` (added a while back to handle ack timeouts) had no guard against you having already rejoined the same channel, so it could fire a stale leave for a channel you were back in - kicking you out server-side while your client still showed `Voice Connected`. That's the screenshot you posted exactly: voice bar says connected, panel says no one is here. Retry now bails out if you've rejoined.

If you hit it again on 3.14.1 let me know - shouldn't need the leave/rejoin dance anymore but I want to hear if anything similar still slips through.

apotheos:
I updated to 3.14.1 and I am running into an issue like this still. It seems the list of connected users doesnt refresh until you refresh the client/website entirely.

ancsemi:
Yeah, the 3.14.1 fix only handled the rejoin-after-disconnect path. The roster-refresh problem you're describing was a separate bug underneath it.

[...]

PR #5521, opened 2026-08-22
Fix channel code rotation state

bernardokcosta:
## Summary

This PR makes server-side channel code rotation consistent across database references, Socket.IO rooms, voice/media state, and pending cleanup timers.

## Changes

- Centralizes manual, join-based, and timed rotations in a shared server-side flow.
- Generates collision-safe codes across channels, server/vanity codes, and managed invites.
- Atomically updates the channel code, AFK references, user preferences, and the AutoMod log channel.
- Migrates active text and voice rooms without disconnecting users.
- Migrates voice, music, screen sharing, webcam, stream viewer, and pending timer state.
- Keeps voice and temporary-channel grace periods valid across rotations.
- Makes temporary-channel deletion transactional and safe against partial failures.
- Uses stable channel IDs for delayed temporary-channel cleanup.
- Adds coverage for rotation persistence, runtime migration, pending timers, and rollback behavior.

## Scope

This PR contains only the server-side rotation work. It does not add code history, permanent aliases, client reconnect gating, or `use_voice` behavior changes.

Client-side reconnect repair is handled separately in #5525.

## Testing

- `node --check src/channelRotation.js`
- `node --check src/socketHandlers/index.js`
- `node --check src/socketHandlers/voice.js`
- `git diff --check`

All 21 tests pass.

### DIFF
--- a/public/app.html
+++ b/public/app.html
@@ -590,6 +590,12 @@
         <div class="right-sidebar-users" id="right-sidebar-users">
           <div class="panel">
             <h5 class="panel-title"><span class="title-emoji">👥 </span><span data-i18n="right_sidebar.online_title">Online</span></h5>
+            <div class="user-search-wrap" id="user-search-wrap" style="display:none">
+              <input type="search" id="user-search" class="user-search-input" autocomplete="off" spellcheck="false"
+                     data-i18n-placeholder="right_sidebar.search_members" placeholder="Search members...">
+              <button type="button" class="user-search-clear" id="user-search-clear" style="display:none"
+                      data-i18n-title="right_sidebar.search_clear" title="Clear">&times;</button>
+            </div>
             <div id="online-users" class="user-list">
               <p class="muted-text" data-i18n="right_sidebar.select_channel">Select a channel</p>
             </div>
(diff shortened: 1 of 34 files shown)


ancsemi:
Thanks for digging into this. The state migration work is thorough and having tests around it is genuinely nice.

I'm hesitant to take it as one piece though. The reconnect path is the most fragile part of Haven by a wide margin, most of my worst bugs have lived there (#5347, #5427, #5391, #5444), and this moves all of the recovery behind waiting for `channels-list` to come back. Today `enter-channel` and `voice-rejoin` fire straight away on reconnect. Normally that extra round trip is nothing, but the conditions those bugs turned up in were exactly the flaky ones where it might not be, and someone sitting in a call is the person who feels it.

Couple of other things I'd rather look at on their own: `channel_code_history` never gets pruned and permanently reserves every code it stores, and old codes resolving to the live channel indefinitely does soften what rotation is for. The new `use_voice` check on `voice-rejoin` is probably the right call, but it's a behavior change that deserves to land by itself so it's easy to back out.

Would you be up for splitting it? The server side (the shared rotation flow, the collision-safe generator, the state migration in `channelRotation.js`) reads well to me and I'd take that part as is. The client reconnect restructuring is the half I'd like to hold and look at separately.

bernardokcosta:
Absolutely, and thanks for the detailed context. I agree that reconnect should not be gated on `channels-list`, especially given the conditions behind #5347, #5427, #5391, and #5444.

I'll narrow #5521 to the server-side work you called out:

- the shared rotation flow;
- collision-safe code generation;
- atomic persistence updates;
- live Socket.IO room and server-state migration;
- the server-side tests.

I'll remove the client reconnect restructuring and restore the current immediate `enter-channel` / `voice-rejoin` behavior.

I'll also remove `channel_code_history` and the indefinite old-code resolution rather than trying to adjust retention in this PR. The new `use_voice` check on `voice-rejoin` will be left for a separate, focused PR as well.

For the reconnect follow-up, I'll preserve the immediate fast path and treat `channels-list` only as a parallel repair/fallback mechanism, with tests for delayed or missing responses.

Thanks — splitting it this way makes sense.

ancsemi:
Merged. This is the split I asked for on #5517 and it's much easier to say yes to on its own.

Rotating a live channel and watching the code change without dropping anyone out of it worked cleanly. The details I liked: refusing to rotate onto the same code, checking the update actually hit one row so a concurrent rotation can't be clobbered, and checking the invite and vanity namespaces for collisions rather than just the channels table. The grace eviction comparing socket IDs before firing is the kind of thing that only shows up under a reconnect race, so good that it's handled.

Eight tests covering the rotation paths made this a lot quicker to review than it would otherwise have been.

PR #5522, opened 2026-08-22
Automatic language selection

bernardokcosta:
## Summary

This PR improves Haven's language selection and fixes inconsistent locale switching across the login page and the main app.

It also replaces OS-dependent flag emoji with bundled SVG artwork, ensuring the Brazilian flag and other language flags render correctly on Linux, Windows, and systems without color emoji fonts.

## Changes

- Adds an **Automatic** language option to the login page and user settings.
- Automatic mode uses:
  1. The server-configured default language.
  2. The first supported browser/system language.
  3. English as the final fallback.
- Reacts to browser language changes while Automatic mode is enabled.
- Preserves explicit user language preferences.
- Prevents slower locale requests from overwriting newer selections.
- Persists the selected preference before reloading the interface.
- Uses a shared language picker across the login page and main app.
- Replaces native flag emoji with bundled SVG flags.
- Adds the Automatic label to all supported locale files.
- Adds accessible keyboard navigation with visible focus, arrow keys, Home, End, and Escape.
- Prevents picker keyboard events from triggering global app shortcuts or closing the settings modal.

## Compatibility

Existing explicit language preferences remain unchanged. Users who want language detection based on server/browser settings can select **Automatic**.

## Testing

```text
node --test test/*.test.js
13 tests passed
```

Additional validation:

- All locale JSON files passed `scripts/validate-locales.js`.
- Tested language switching between Automatic, English, and Portuguese.
- Tested persistence and page reload behavior.
- Tested the login page and main app on desktop and a `390×844` mobile viewport.

### DIFF
--- a/public/app.html
+++ b/public/app.html
@@ -1210,14 +1210,15 @@
         <label class="select-row">
           <span data-i18n="settings.language_section.label">Language</span>
           <select id="language-select" class="settings-select">
-            <option value="en">🇬🇧 English</option>
-            <option value="fr">🇫🇷 Français</option>
-            <option value="de">🇩🇪 Deutsch</option>
-            <option value="es">🇪🇸 Español</option>
-            <option value="pl">🇵🇱 Polski</option>
-            <option value="ru">🇷🇺 Русский</option>
-            <option value="zh">🇨🇳 中文</option>
-            <option value="pt">🇧🇷 Português (Brasil)</option>
+            <option value="auto" data-i18n="settings.language_section.auto">Automatic</option>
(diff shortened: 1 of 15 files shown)


ancsemi:
Merged, thanks for this one. The auto-detected language getting saved as if it were an explicit pick was a real problem, it meant the server's default language setting never reached anyone who had already visited once. Bundling the flag SVGs was a good call too, the emoji ones render as plain letters on a lot of setups.

Tests and the locale validator both pass on main.

PR #5523, opened 2026-08-23
[feat] Navidrome rich presence, live-updating profile card

Bo0sted:
Shows what a user is playing in Navidrome on their Haven profile, via the [Havidrome](https://github.com/Bo0sted/HavidromePlugin) companion plugin, and makes the profile card update live from presence while it's open.

## What's included

**Navidrome rich presence**
- Users generate a revocable webhook token in Settings > Activity > Navidrome.
  Only the token reaches the client; it builds the full URL from its own origin,
  so the server never needs to know its public address.
- The plugin posts strict multipart presence to a reserved per-user path
  (`/api/webhooks/navidrome/:token`): title and duration required, optional
  cover bytes.
- Cover art is magic-byte checked and held in Haven's memory only, then served
  to viewers from `/api/activity/navidrome-cover/:userId`. Navidrome itself is
  never contacted and its address never leaves the plugin.
- Presence runs through the existing activity slot, so it stays online-gated,
  respects the user's music-sharing preference, clears on disconnect, and auto
  expires at track end. A paused track expires after 5 minutes idle so it can't
  linger and reappear on reconnect.
- The profile card gains a live progress bar: pause freezes it, resume continues
  it, a new track swaps it, stop clears it.

**Live profile card**
- The open profile card previously froze at whatever it showed when opened. It
  now re-renders status (online/away/dnd/offline), custom status text, and
  activity in place from the online-users broadcasts the client already
  receives. No new traffic.
- Fixes the card clipping off the right edge: positioning used a hardcoded
  320x400 guess, but the card's width is a rem value and its height varies with
  content, so it spilled once font-size or zoom differed from 16px. It now
  measures the real rendered box before clamping.

## Security and privacy

- 64-char hex token, validated on every request; parameterized lookups.
- Rate limiters on both the webhook and the cover route, plus upload size and
  field caps on the webhook.
- Cover bytes are accepted only if their magic bytes are a real jpeg/png/gif/webp
  (SVG rejected, so no script can hide in an "image").
- Each cover URL carries a hash of that image. The server serves the bytes only
  when the hash matches the current cover, which refreshes the art on a track
  change and stops covers being fetched by guessing user IDs.
- Presence is filtered by the existing share_music_activity preference at read
  time, so it inherits Haven's privacy model.

## Translations

Adds the `settings.navidrome.*` keys to `en.json` (main). Other locales fall
back to English until community translators fill them in;
`scripts/validate-locales.js` passes for all locales.

## Testing

- `node scripts/validate-locales.js` passes (all 7 locales).
- Verified live updates, pause/resume/track-change/stop, offline clearing, the
  paused-idle expiry, and the off-screen positioning fix.


ancsemi:
Read the whole thing. Nothing worrying on the security side: the token handling, the magic-byte check on covers and the version hash on the cover URL all hold up, and Navidrome itself never being contacted is a nice property.

Could you split it though? There are three things in here and two stand on their own. The live profile card and the off-screen positioning fix are bugs that hit everyone whether or not they run Navidrome, and I'd like to take those now. The Navidrome integration is a bigger decision for me (new table, a new unauthenticated webhook route, and it leans on a companion plugin), so I want to sit with that one separately rather than have it ride along.

Where the line falls: the progress bar goes with the Navidrome half. Nothing else on main sets `duration` on the listening slot, so `_activityProgressHtml`, `_startActivityProgress`, `_formatClock`, the `.pap-*` CSS and the paused verb are all inert without it.

Two notes on the fixes themselves.

The live refresh rewrites the activity and status slots with innerHTML on every `online-users` broadcast. That fires on joins, leaves, status changes and every activity poll, not only when something about the profiled user actually changed, so a card showing cover art recreates its `` each time. Keeping the last rendered HTML and skipping the write when it matches would settle that.

The card also can't flip to offline, and which way that goes depends on a server setting. With member visibility on "all" the list carries offline members with `online: false` and it works. On the default the user drops out of the list entirely, your `if (!u) return` guard correctly leaves the card alone, and it keeps showing them online for as long as it stays open. Leaving it alone beats wrongly flipping, so I'd keep the behaviour, just say so in the comment so the next person reads it as deliberate.

Positioning I'm happy with as written. I checked that the entrance animation is transform only and that the avatar, banner and activity art all have fixed sizes, so `offsetHeight` straight after append really is the final height. One thing you could pick up while you're in there: View Full Bio grows the card after it's been positioned, so one sitting near the bottom overflows once expanded. That predates your change, but now there's an accurate measurement to re-run.

Two small ones on the Navidrome side so they aren't a surprise when it comes back around. Multer buffers the upload before the token is looked up, so an unknown token still costs up to 512KB; checking the token first and running the upload after would tighten it. And `duration` has no upper bound, which bites because a large enough value overflows setTimeout's 32 bit range and the expiry timer fires immediately instead of at track end.

Bo0sted:
Updated to be generic and compatible with any music server that supports plugins / custom scripts :)

ancsemi:
Merged, thanks. The generic version is exactly what I was hoping for, and the doc reads well.

Three small things I did on top, in 8429e14: the webhook now resolves the token before the upload parser runs, so a post to a bad token doesn't buffer 512KB first; the open profile card compares against what it last rendered before rewriting its slots, since online-users fires on every join and poll and the cover image was being recreated each time; and the API doc moved to docs/listening-api.md next to the other guides, with the Settings link updated to match.

PR #5524, opened 2026-08-24
More Invite Refinements

birdcrazy:
### Added a "used" status to invite link cards with it's own status color.
An invite that has had all of it's uses redeemed, displays as "Active" until it expires.
The new "used" status makes it easier to see at a glance which links are actually still redeemable.

### Increased the width of the "expires" dropdown in the invite link creator and editor
When the selected option was anything other than "Never", the selected text was overflowing the dropdown field.
Now the dropdown is widened, so that the longest text option is fully visible in the field when selected

### New Setting: max-invite-uses
Any user with invite-users permissions could create unlimited use invite links and potentially post them in public spaces, which essentially would make the server publicly available, even if the admin doesn't desire this. 

The new max-invite-uses setting allows the admin to control how many uses a user can create an invite link for. (admins and server-managers ignore this limit)

Added the new setting to the admin settings (defaults to 1), as well as AI translated keys for the setting title and description.

### Fix invite update toast
When editing a invite link, if the server rejected the edit for any reason (such as invalid values), both the "success" and "error" toast messages would display.
Now this behavior is fixed, so that the success toast only shows when the link update was actually successful. 

_Related to #5470_


ancsemi:
The Used status and the wider expiry dropdown are both things that have bugged me, glad to see them. Found a few things before I merge though.

The max uses limit can be worked around. The check runs on the raw payload, so if `maxUses` just isn't sent at all, both `data.maxUses > maxInvtUsesSetting` and `data.maxUses <= 0` are false against `undefined`, it passes, and the line below coerces it to 0 (unlimited). A non-numeric value does the same thing. Validating after the coercion instead of before would close it.

Two smaller ones:

`ic.use_count >= ic.max_uses` marks unlimited links as Used. `max_uses` 0 means unlimited, so `0 >= 0` is always true and every unlimited link shows up as spent.

And `value="${ic.max_uses || 1}"` in the editor renders 1 for an unlimited link. The save handler always sends `maxUses`, so editing just the label on an unlimited invite quietly turns it into a single use link.

One question rather than a bug: defaulting `max_invite_uses` to 1 caps delegated inviters on existing servers the moment they update. I lean toward 0 as the default so nothing changes for people unless they opt in, but open to your thinking there.

Happy to merge once those are sorted.

birdcrazy:
Good Catches!
I made changes to to fix the issues you pointed out:
- Changed default max_invite_uses setting to 0
- Fixed `maxUses` omition/wrong-value bypassing `maxInvtUsesSetting` in both `socket.on('create-invite-code')` and `socket.on('update-invite-code')`
- Moved the `maxUses` checks for `socket.on('update-invite-code')` inside `if ('maxUses' in data) {` to allow partial updates instead of just failing the invite update when `maxUses` is not provided.
- Added new error message emit when maxUses > 100000 instead of silently limiting the value.
- Fixed false "Used" status for infinite use invite links.
- Changed `value="${ic.max_uses || 1}"` -> `value="${ic.max_uses || 0}"` to fix infinite link turning into a single use link.

The intent behind defaulting `max_invite_uses` to 1 was to make the default restrictive, then the admin could loosen it up if desired. But You are right that migrating existing deployments to behave as they currently are, so the behavior of exiting setups doesn't change silently, is probably more important right now.

ancsemi:
Merged, thanks! The cap is the part I looked at hardest, since a setting that only narrows the input box would be trivial to walk around. It's enforced server side on both create and edit, and admins and manage_server bypass it as documented, so it holds up.

Requiring maxUses to actually be a valid number rather than quietly treating a missing value as unlimited is a good catch too. That's the sort of default that bites much later.

Also nice that you moved the "Invite link updated" toast to the server. The old one fired whether or not the update went through.

PR #5525, opened 2026-08-25
Repair client state after channel code rotation

bernardokcosta:
## Summary

- Detect channel-code rotations by stable channel ID and migrate runtime and persisted client state.
- Preserve immediate `enter-channel` and `voice-rejoin` emissions during reconnects.
- Repair stale channel, voice, pin, thread, mute, hidden-channel, and category state after `channels-list` arrives.
- Make voice recovery race-safe with single-flight joins and a bounded 6-second `voice-channel-gone` deferral.
- Scope persisted channel mappings to the authenticated user and clear them during logout or session invalidation.
- Keep the persisted map updated across consecutive live rotations.

This complements the server-side state migration in #5521 while remaining independently based on `main`.

## Testing

- `node --check public/js/modules/app-socket.js`
- `node --check public/js/modules/app-ui.js`
- `node --check public/js/voice.js`
- `node --check test/channelCodeClientState.test.js`
- `git diff --check upstream/main...HEAD`

### DIFF
--- a/public/js/modules/app-socket.js
+++ b/public/js/modules/app-socket.js
@@ -837,6 +837,7 @@ _setupSocketListeners() {
       this.socket.emit('voice-rejoin', { code: voiceRotation.newCode });
       if (this.voice.isMuted) this.socket.emit('voice-mute-state', { code: voiceRotation.newCode, muted: true });
       if (this.voice.isDeafened) this.socket.emit('voice-deafen-state', { code: voiceRotation.newCode, deafened: true });
+      this.voice._healPeerConnectionsAfterChannelRotation?.(voiceRotation.oldCode);
     }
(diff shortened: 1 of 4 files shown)


ancsemi:
Merged on top of #5521.

Checked the piece I was most wary of, the persisted code map, since channel codes are effectively invite tokens and a stale one sitting in localStorage after someone else signs in on the same browser would be a real leak. Scoping it by owner id and returning an empty map on a mismatch is the right call. I confirmed a live rotation updates the stored map and the client follows the new code without losing its place.

The single-flight voice join and the bounded deferral on voice-channel-gone are the parts I'd have expected to get wrong, so it's good to see tests pinning both.

PR #5531, opened 2026-08-26
Add secure bot voice gateway

bernardokcosta:
## Summary

Adds an opt-in voice gateway that allows webhook bots to join Haven voice channels as external WebRTC peers.

## Changes

- Adds `webhooks.can_use_voice`, disabled by default and configurable only by admins.
- Authenticates bots through `socket.handshake.auth.botToken`.
- Uses negative synthetic user IDs to avoid collisions with regular users.
- Adds `GET /api/webhooks/:token/voice/channels`.
- Supports voice join/leave, offer/answer/ICE signaling, speaking state, and AFK activity.
- Displays connected bots in voice rosters with a `BOT` badge.
- Revalidates access per packet and every two seconds.
- Immediately disconnects bots when permission is revoked, the webhook is disabled/deleted, or its channel scope changes.
- Prevents duplicate bot processes and rejects signaling from replaced sockets.
- Keeps bot state synchronized during channel code rotation.
- Prevents bots from joining text/admin rooms or receiving global broadcasts.
- Hides webhook tokens from non-admin integration managers who do not own the webhook.

## Security

- Voice access remains disabled for all existing webhooks.
- Bots can access:
  - Their assigned channel.
  - Channels their creator belongs to.
  - Every non-DM channel when owned by an admin.
- Voice-disabled and full channels reject joins without removing the bot from its current channel.
- Signaling is accepted only from the socket that currently owns the bot's voice entry.
- Access is periodically revalidated to cover out-of-band permission and membership changes.

## Testing

- Added tests for channel scope, membership, admin ownership, and voice-disabled channels.
- Added tests for permission and token revocation.
- Added tests for authorized signaling and replaced socket rejection.
- Added tests for duplicate processes, channel limits, and AFK activity.
- Added tests for BOT presence, room isolation, global broadcast isolation, and channel rotation.
- Added an end-to-end test for the REST endpoint using a real server process.
- `node --test test/*.test.js`: **42 passed**
- Relevant `node --check` validations passed.
- `git diff --check` passed.

### DIFF
--- a/public/js/modules/app-channels.js
+++ b/public/js/modules/app-channels.js
@@ -1685,7 +1685,9 @@ _renderWebhookList(webhooks, channelCode) {
     return;
   }
   container.innerHTML = webhooks.map(wh => {
-    const maskedToken = wh.token.slice(0, 8) + '••••••••';
+    const maskedToken = typeof wh.token === 'string' && wh.token
+      ? wh.token.slice(0, 8) + '••••••••'
+      : 'Hidden - owner/admin only';
     const statusLabel = wh.is_active ? `🟢 ${t('channels.webhook_active')}` : `🔴 ${t('channels.webhook_disabled')}`;
     const toggleLabel = wh.is_active ? t('channels.webhook_disable') : t('channels.webhook_enable');
     return `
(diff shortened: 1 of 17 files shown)


ancsemi:
Merged. Thanks for trimming this down from the first pass, dropping the Node bump, the extra dependency and the soundboard changes made it a much easier read.

For the record it went in after the search PR and merged clean, and the full test suite passes on main including your 15 new ones.

PR #5532, opened 2026-08-26
Add RepoCloud deploy button

cosark:
This PR adds RepoCloud as a one-click cloud deployment option alongside the existing Zeabur button.


ancsemi:
Merged, thanks. Zeabur was already sitting there, so a second click to deploy option fits fine.

One thing worth saying since the template lives on your side: if it ever drifts behind current Haven, I'll pull the button until it catches up. Nothing for you to do right now.

PR #5533, opened 2026-08-26
[big feat] Search overhaul

Bo0sted:
## Search overhaul

Reworks message search end to end: a redesigned persistent search panel, real global search backed by SQLite FTS5, a full filter set with dropdown pickers, and results that render as real messages. Everything is built to reuse the existing chat pipeline and keep the server authoritative.

### Panel and UX
- Search is now a Discord style overlay panel on the right, identical in public channels and DMs. It only truly closes on the X; navigation just hides and shows it, preserving query, results, page, and scroll.
- Per context state: one shared panel across all public channels, one independent panel per DM.
- Numbered pager, 25 results per page, with a page number input so you can jump straight to a page instead of clicking arrows one fetch at a time.
- Stale response guard (tokened requests), a loading state, and recent searches.

### Real search (FTS5)
- Global search across every non DM channel the user belongs to, backed by SQLite FTS5 instead of the old per channel `LIKE` scan.
- Access scoped in SQL to the user's channel memberships; `in:#channel` narrows and re-checks membership. Server paginates with `LIMIT/OFFSET` plus a `count(*)` total.
- Sort by newest, oldest, or most relevant (bm25).
- Tokenizer is admin configurable (`HAVEN_SEARCH_TOKENIZER`, default `unicode61` for Discord like whole word and prefix matching; `trigram` and `porter` available). Boot time reconcile detects the live tokenizer and rebuilds the index only when it changes, inside a guarded transaction so an interrupted switch rolls back and retries.

### Filters and dropdowns
- Filter popover under the search box (public channels only): `from:`, `in:`, `has:`, plus `pinned:true` and date filters `before:` / `after:` / `during:`.
- `has:` covers image, video, audio, file, and link. Dropdown lists come from client state only, since the server re-authorizes every query anyway.
- Each result shows its channel, and clicking a result jumps into that channel and to the message.

### Results as real messages
- Results render through the message formatter, so images, uploaded video and audio, spoilers, mentions, and links look the same as in chat. Audio uses `preload="none"` so nothing loads until play.
- Right clicking a result gives Copy link and Delete. Delete targets the result's own channel and the row is removed only after the server confirms the delete.
- Results that head a thread show a reply count badge.

### Access and safety
- **Server stays authoritative.** Every query re-checks channel membership; delete re-authorizes against the message's real channel. The client is convenience, not the gate.
- **Embeds are curbed behind a Load button.** A page of 25 results never fires a burst of preview fetches. Each page link gets a Load button that resolves just that one embed through the existing scheduler and shared caches. One additive skip in the preview pass; channel and DM views are unchanged.
- **Per account rate limit.** Search gets its own bucket (10 requests per 10 seconds), keyed per account like the rest of Haven, since an FTS match plus count plus cross channel joins is heavier than the shared event bucket assumes. Sliding window, so blocked attempts do not extend the cooldown; input is debounced so normal use never trips it.
- **Live permission invalidation.** The panel cache reacts to the existing `channels-list` and per user `roles-updated` events and shows a refresh banner, so a ban, kick, role change, or channel removal cannot leave stale results visible. No new server plumbing.
- **DMs stay client side.** DMs are E2E encrypted, so DM search runs locally over decrypted messages and never touches FTS.
- **No global side effects.** Search rows carry `.message-content`, but every message pass is container scoped.


ancsemi:
Merged, thanks. Scoping access in SQL and rechecking it on every query is what made this an easy yes, and keeping DMs client side was the right call.

One note for anyone updating a server with a lot of history: the first boot after this lands does a full index rebuild, so search can look empty for a minute before it settles.

Issue #5426, opened 2026-06-21
Audio desync and stutter + audio duplication when screen sharing


RCCore:
Haven version: v3.25.2;
Client version: Windows desktop v1.4.24

The problem is when I try to share my screen, the video looks fine, but the audio starts to stutter and loses sync with the video. I'm using Coturn as my STUN & TURN server and sometimes the audio duplicate.

ancsemi:
I found one cause of the audio doubling: the volume control was leaving the stream playing through two outputs at once after you changed volume or closed and reopened a stream, so you'd hear it twice. That's fixed in the latest commit on main and will go out in the next build.

A few questions on the desync + stutter when you get a chance:

- Is it the people watching your screen who get the stutter, or do you hear it on your end too?
- How many people are in the call when it happens?
- Does it still happen when you and someone else are on the same local network, or only when the connection routes through your TURN server?
- In Settings > Debug, is "Apply voice processing to screen-share audio" on or off?

Once I know whether it's relay-only and who's affected, I'll have a much better idea where to look.

RCCore:
1. For the first question yes it is the people watching my screen and when somebody else share the screen can I hear it too.
2. For the second one it can be only one or two people or many more when it happens.
3. For the third one it works well on LAN. (came to think of it maybe it is my TURN server but I am not sure because the video plays fine only the audio is bad).
4. And the last one I have the option off.

Oh and another problem I have is when somebody share the screen once I close that stream I cannot enter it without rejoining the call or restarting the app. (I use only the app not the browser).

RCCore:
I have another bug to report. When I add a custom soundboard, sticker or emoji to the app, these changes are not visible to other users. The app needs to be restarted for the changes to take effect for other users, but not for the admin.

ancsemi:
Got the soundboard/sticker/emoji one fixed. Those get uploaded over a separate channel that didn't tell anyone else about the change, so only the person who added it saw it until everyone else restarted. Now the server pings all connected clients to refresh that list the moment something is added or removed, so it shows up live. That's on main for the next build.

Still digging on the screen-share audio desync/stutter over your TURN server, that one's trickier since it only shows up on relay and not LAN. The duplicate-audio part should already be handled in the latest build. And the "can't reopen a stream after closing it" thing you mentioned should be sorted in 3.26/3.27 (you were on 3.25.2), so give that another go once you update and let me know if it still happens.

RCCore:
[...]

Issue #47, opened 2026-02-21
Voice calls and screen sharing only work for 2 people


USBAkimbo:
Hello, me again lol

So I've been doing some more testing with Haven and I got a few people to join my server and experienced this

- I join voice
- Friend 1 joins voice
- We can talk and he can see my screen share
- Friend 2 joins voice
- Me and friend 1 can still talk and see the screen share, but we cannot hear friend 2
- Friend 2 cannot hear either of us and cannot see the screen share

I'm using coturn for this setup as well

I also tested joining from my phone on 5G and experienced the same - nobody could hear me and I couldn't hear them

However, image/file sharing works, so does typing indicators and messages

We also noticed that if 2 people are talking in a channel, 2 people cannot call in a private channel

ancsemi:
hey! thanks for the detailed report, this was really helpful.

so the mesh signaling itself actually supports N users just fine — the real issue was a bug where the TURN credentials were never being fetched because of a wrong localStorage key in the voice client. that meant only direct STUN connections worked, which explains why it'd work for some pairs on LAN but fail for others (especially across networks or on 5G).

that bug was fixed in v2.2.1 which just went out. if you pull the latest and restart, TURN should actually kick in now and all your users should be able to connect properly — including the phone-on-5G case.

the "2 people max per voice channel" thing should also go away since it was the same underlying connectivity issue. let me know if it's still happening after the update!

USBAkimbo:
@ancsemi https://github.com/ancsemi/Haven/actions/runs/22214765175

Looks like the Docker image build is failing - my local build of it in compose is failing too

Turns out I was running a slightly older version due to a config issue on my end, so as soon as that image is available I can test it

ancsemi:
Good catch on the Docker build, thanks for flagging it. You're right that the QEMU ARM emulation is crashing on native modules like better-sqlite3. Your native ARM runner approach is the right call — I just need to add a merge step so both platforms end up in one manifest list instead of overwriting each other's tags. Will get that sorted out.

USBAkimbo:
@ancsemi https://github.com/ancsemi/Haven/actions/runs/22262049905/job/64401802195

We're almost there

```
Error: buildx failed with: ERROR: failed to build: invalid tag "ghcr.io/ancsemi/Haven:build-amd64": repository name must be lowercase
```

Just a typo that needs fixing lol

ancsemi:
[...]

Issue #45, opened 2026-02-20
coturn has stopped working since 2.0.1


USBAkimbo:
So I was running 2.0.1 with coturn and logged on today to find that it doesn't appear to be working anymore - calls and livestreams are just black

I updated to 2.2.0 and I'm still having the same issue - my calls and streams over the internet aren't working

I've checked my Oracle Cloud firewall rules and all the ports for coturn are still there

I'm building my infra using Ansible too so I don't believe it's a config issue

Here's my compose for example

I'm not sure what's wrong - has there been an update that changed how coturn works with the app?

```
services:
  haven:
    image: haven:2.2.0
    build:
      context: https://github.com/ancsemi/Haven.git#v2.2.0
    container_name: haven
    restart: unless-stopped
    volumes:
      - {{ DOCKER_VOLUMES_ROOT }}/haven/data:/data
    environment:
      - PORT=3000
      - HOST=0.0.0.0
      - SERVER_NAME=Akimbo's
      - ADMIN_USERNAME=akimbo
      - TURN_URL=turn:{{ HAVEN_URL }}:3478
      - TURN_SECRET={{ TURN_SECRET }}
    networks:
      - proxy
    labels:
      - traefik.enable=true
      - traefik.docker.network=proxy
      - traefik.http.routers.haven.rule=Host(`{{ HAVEN_URL }}`)
      - traefik.http.routers.haven.entrypoints=websecure
      - traefik.http.routers.haven.tls.certresolver={{ TRAEFIK_CERT_RESOLVER_NAME }}
      - traefik.http.services.haven.loadbalancer.server.port=3000
      - traefik.http.services.haven.loadbalancer.server.scheme=https
      - traefik.http.routers.haven.service=haven

  coturn:
    image: coturn/coturn:4.8.0-alpine
    container_name: coturn
    restart: unless-stopped
    network_mode: host
    command:
      - -n
      - --listening-port=3478
      - --tls-listening-port=5349
      - --realm={{ HAVEN_URL }}
      - --use-auth-secret
      - --static-auth-secret={{ TURN_SECRET }}
      - --min-port=49152
      - --max-port=65535

networks:
  proxy:
    external: true
```

ancsemi:
hey, good catch — this one was completely my fault. turns out the TURN feature was broken from day one and i never noticed because i was always testing on LAN.

[...]

PR #5535, opened 2026-08-26
Add native screen sharing signaling transport

bernardokcosta:
## Summary

- Add authorized Socket.IO signaling for native screen-share offers, answers, and ICE candidates.
- Track screen-share transport, session, and negotiation identifiers to reject stale signaling.
- Preserve and reannounce active shares across voice reconnects.
- Add bounded offer rate limiting and pending ICE handling.
- Clean native screen-share state during channel deletion, rotation, expiration, moderation, and disconnect flows.
- Keep the existing browser screen-sharing transport fully supported.

## Testing

- 55 tests passed.
- `node --check` passed for all changed JavaScript files.
- `git diff --check` passed.

## Paired PR

Desktop implementation: https://github.com/ancsemi/Haven-Desktop/pull/47


ancsemi:
Read through this one along with the Desktop half. The signalling is careful, the session/negotiation IDs for rejecting stale offers make sense, and I like that the browser transport stays intact.

Holding it briefly rather than merging today, for one reason. #5426 is an open screen-share audio regression that the reporter is about to retest on the current build, and this touches the same paths in voice.js that the fix went into. If it lands first and he reports it still stuttering, there's no way to tell which change to look at.

Nothing wrong with the PR. It's sequencing, and it also deserves a proper look on its own rather than being reviewed alongside six other things, given it's roughly 4700 lines across both repos.

I'll come back to it once #5426 is confirmed one way or the other. If that goes quiet for a week I'll pick this up anyway rather than leave it sitting.

bernardokcosta:
One potentially relevant data point for #5426: I've been testing a local Linux Desktop build with my other open Desktop changes, including Haven-Desktop#45, but **without #47 and without the paired #5535 native screen transport**. So the native H.264 path is not involved in this test.

On my previous Linux setup I was consistently seeing two screen-share audio problems: Haven's own call audio could get captured back into the screen share, and the shared audio could accumulate noticeable latency/desync.

#45 specifically targets those parts of the Desktop audio path: it excludes Haven and its child processes from system capture, removes the unsafe raw Electron loopback fallback, bounds the native-to-JS and renderer PCM queues, reduces the Linux capture buffering, and drops stale PCM instead of allowing latency to accumulate.

With the integrated Desktop build I'm using now, screen-share audio has been stable for me so far and I'm no longer reproducing either of those problems.

I don't want to claim that this proves #45 fixes #5426. RCCore is testing Windows + Coturn, my test is Linux, and my build contains several Desktop PRs, so this is not a clean #45-only A/B test. #5426 also clearly has separate fixes in `voice.js` around WebRTC timing, audio priority, duplicate negotiation and recovery.

Still, RCCore previously mentioned that the audio behaves better in the browser than in the Desktop app, so I think #45 may be relevant to a Desktop-specific part of the regression.

Also, since #47 is currently video-only and isn't present in the build I'm testing, the improvement I'm seeing cannot be coming from the new native H.264 screen transport.

I'm fine with holding #5535 until the current-build #5426 retest is done; I just wanted to add this data point in case it helps separate the Desktop capture path from the signaling/WebRTC side of the issue.

ancsemi:
Quick heads up, 5531 just landed, so this one now conflicts on voice.js. No rush while it's parked, but it'll need a rebase whenever we pick it back up.

bernardokcosta:
> Quick heads up, 5531 just landed, so this one now conflicts on voice.js. No rush while it's parked, but it'll need a rebase whenever we pick it back up.

Conflicts are resolved now by merging the latest `upstream/main` into this branch.

ancsemi:
Read the whole thing and ran it, 133/133 pass, and it merges clean. The signaling handler is careful in the places that matter: it checks the sender is actually in the voice room on the socket it claims, that the target exists and is not the sender, that a native session with that id is really live, and it rate limits offers per target. No new dependencies, and you picked up the SDP cap sitting under socket.io's frame limit rather than on it, which I only fixed on main this morning.

I am not merging it yet, and the reason is timing rather than the code.

Two things are in flight in exactly this subsystem. #5543 landed a few hours ago and reworked offer and answer correlation, and there is an open report in #5542 of web to web voice failing that I have spent this evening chasing. I reproduced two browser clients negotiating and passing media on current main, so I do not think we broke it, but that thread is not closed. This PR is 765 lines into `voice.js` and it does touch the existing browser path, not only add a native one alongside it, so landing it now would make it very hard to tell which change caused what if something surfaces next week.

The other thing is the Desktop half. ancsemi/Haven-Desktop#47 has not been touched since the 26th while this side has moved, so I would want the two rechecked against each other before either goes in, given neither is much use alone.

What I would suggest, and it is genuinely a question rather than a decision I have made: give #5542 and #5426 a few days to either settle or turn into something, then land this and the Desktop PR together with the pair verified against each other. If you would rather have it in sooner, say so and I will do a full pass on the browser-path changes specifically, since that is the part that carries the risk and the part the tests can say least about.

Nothing here needs changing on your end in the meantime.

ancsemi:
Came back to this now that 5426 is confirmed better and 5542 is closed. Read both halves again, and the signaling still looks solid, it merges clean and the suite passes on the merged tree.

The blocker is on the Desktop half. As it stands, whenever the machine supports it the native path is used automatically, it is video only, and it swaps the thumbnail picker for a plain menu. So the moment this lands, Windows desktop users lose screen share audio, including the per app capture. Losing per app audio is a non starter for me.

What would get this in: make the native transport opt in (a Desktop setting, default off) so the existing Chromium path with the audio picker stays the default, and say in the setting that native is video only for now. Once that is agreed I am fine landing this server PR first, since it is backwards compatible, and the Desktop PR after it. One more thing worth a look: bundling GStreamer grows the installer a fair bit, so if there is a way to trim what gets staged, that would help.

PR #5536, opened 2026-08-27
Misc theme picker fixes

birdcrazy:
Fixes some minor inconsistencies with the theme picker(s)

Issues that were fixed:
1. Some of the included themes were not available in the "Default Theme" admin dropdown setting. the options also were not in the same order as the theme picker.
2. The themes that were missing from the "Default Theme" setting were not on the allow list to save as the default theme in the DB.
3. The theme selector in the login-page did not include all the system themes.
4. When a new user navigates the the login page, the admin selected default theme loaded, but the associated theme selector button did not display as "active". The Haven theme button was always the "Active" one.

---
Worth noting that any new, removed, or renamed themes in the future will need the code modified in 4 places.
1. _public\index.html_ under `<div class="auth-theme-bar" id="auth-theme-bar">`
2. _public\app.html_ under `<div class="theme-selector" id="theme-selector">`
3. _public\index.html_ under `<select id="default-theme-select" style="max-width:160px">`
4. _src/socketHandlers/admin.js_ in `const validBuiltin = [`

We could make all of these places dynamically built from one shared list, so new themes only need to be added in a single location. But there are already plenty of default shipped themes that adding more in the future will be rare/unlikely. And it would be slightly heavier on compute than the current hardcoded method.
For these reasons I left them as hardcoded for now. 

Let me know if you want it refactored to share a common theme-list


ancsemi:
Merged, thanks. The active-button one had been bugging me and I hadn't tracked down why.

Two small follow-ups on top: Zelda was missing from the dropdown and the allowlist as well, and I moved the login page back to applying the default without storing it. Storing it means the page never consults the server default again, so an admin changing the default later would never reach anyone who has already loaded the login page and not signed in. `applyThemeFromServer` takes a `persist` flag now, and `applyEffects` takes the theme so the overlays still match when nothing is stored. The app page keeps storing it, since it re-applies the default on every connect anyway.

On the shared theme list: agreed, leave it. Four places is annoying but adding a theme is rare and a single list would have to be readable from both pages and the server, which is more machinery than the problem deserves.

PR #5540, opened 2026-08-28
Add dynamic bot voice audio

bernardokcosta:
## Summary

Adds secure dynamic audio playback for bots connected to Haven voice channels.

Bots can upload MP3, WAV, or OGG files through the webhook API. Haven validates the file, queues it, and synchronizes playback for users currently connected to the same voice channel.

## What changed

- Added audio upload, skip, and stop webhook endpoints.
- Added a global queue per bot with a maximum of 25 items.
- Added file size and duration limits of 10 MiB and 5 minutes.
- Added magic-byte and metadata validation for MP3, WAV, and OGG files.
- Added short-lived capability URLs for private audio delivery.
- Added synchronized playback for late joins and reconnects.
- Added client support for deafen state and output device selection.
- Added cleanup for disconnects, permission revocation, channel deletion, rotation, restore, update, and shutdown.
- Excluded temporary bot audio from static uploads, storage accounting, and backups.
- Updated Multer to 2.2.0 and added `music-metadata`.

## API

- `POST /api/webhooks/:token/audio`
- `POST /api/webhooks/:token/audio/skip`
- `DELETE /api/webhooks/:token/audio/current`
- `GET /api/bot-audio/:playbackId/:accessToken`

Audio operations require the bot to have voice permission and to be actively connected to an accessible voice channel.

## Security

- Authorization and voice presence are revalidated after asynchronous file inspection.
- A final synchronous authorization check runs immediately before enqueueing.
- Playback URLs are random, short-lived, and valid only for the current queue item.
- Static access to the bot audio directory is denied.
- Aborted and rejected uploads are removed.
- Files with spoofed extensions or unsupported MPEG layers are rejected.


ancsemi:
Read all of it and ran it. The full suite passes on your branch, 67 tests including your 665 lines, and the integration test genuinely exercises the thing rather than walking a happy path.

The design holds up. Revalidating the token, the voice permission and the channel after the async inspection, then once more synchronously right before enqueue, is the part I would have worried about and you closed it. Magic bytes plus the container/codec cross-check is a real check rather than trusting the extension. And you found every place bot audio has to be kept out of: static serving, backups, upload accounting, and all three UPLOAD_PATH regexes. The lifecycle cleanup is more thorough than I expected.

One thing has to change before this can go in.

The `/uploads/bot-audio` block copies a pattern that turns out not to work. I planted a file in that folder on your branch and got its bytes back from all three of:

```
/uploads/bot%2Daudio/leak.mp3
/uploads//bot-audio/leak.mp3
/uploads/bot-audio%2Fleak.mp3
```

Express matches a mounted prefix against the raw path while express.static decodes before it resolves, so the two disagree about what is being asked for. For bot audio that is not cosmetic: the point of the capability URL is that the file is not reachable any other way, and it is, by name, with no token.

Not your bug, though. The deleted-attachments guard directly above it had the same hole, and worse, since a file moved there is meant to be unreachable and anyone who saw the attachment while it was posted knows its filename. I fixed that on main in b4ddcb8: the guard now decodes, resolves against the uploads root and checks containment, verified against thirteen spellings. When you rebase, drop your `app.use('/uploads/bot-audio', ...)` line and add `'bot-audio'` to the `BLOCKED_UPLOAD_DIRS` array instead.

Two smaller ones, neither blocking:

`/api/bot-audio/:playbackId/:accessToken` is the only new route without a rate limiter. The format check makes guessing impractical, but every request walks the channel map, so it may as well have one for symmetry.

`req.once('aborted')` and `req.aborted` are both deprecated on IncomingMessage. They work, your abort test proves it, but `res.on('close')` with `req.destroyed` is the shape that survives.

Incidentally the multer bump is not load bearing. I ran everything against 2.0.2 and it all passed, so if 2.2.0 is wanted for its own sake that is fine, it just is not a dependency of this feature.

bernardokcosta:
Rebased onto b4ddcb8 and addressed all review points.

- Removed the ineffective mounted /uploads/bot-audio prefix guard and added bot-audio to the decoded, resolved BLOCKED_UPLOAD_DIRS containment check.
- Added regression coverage using a real existing audio file for the direct path, encoded hyphen, duplicate slash, and encoded separator spellings.
- Added a separate rate limiter to the capability playback route.
- Replaced the deprecated IncomingMessage aborted event/property flow with response close handling, req.destroyed, and a synchronous disconnect flag before enqueue.
- Updated Multer to 2.3.0. You were right that the bump is not load-bearing for this feature; 2.3.0 is retained independently because it fixes the 2.2.0 aborted-upload file descriptor leak (GHSA-qfvm-cv95-jqjf).

The focused suite passes 30/30 and the full clean suite passes 73/73.

PR #5541, opened 2026-08-28
Add webhook bulk message deletion

bernardokcosta:
## Summary

Adds a moderation-protected webhook endpoint for deleting recent messages in bulk from a bot's assigned channel.

## API

```http
DELETE /api/webhooks/:token/messages?limit=25

### DIFF
--- a/GUIDE.md
+++ b/GUIDE.md
@@ -1321,6 +1321,17 @@ DELETE https://your-server.com/api/webhooks/<token>/messages/<message_id>
 
 Bots can delete any message in their assigned channel. Returns `{ "success": true }`.
 
+Bots with moderation permission can delete up to 100 of the most recent messages
+from their assigned channel in one request. Thread replies are removed with their
+parent messages.
+
+```
+DELETE https://your-server.com/api/webhooks/<token>/messages?limit=25
+```
+
+Returns `{ "success": true, "deleted": 25 }`. The `deleted` count includes
+thread replies removed with the selected messages.
+
 ### Playing Soundboard Sounds
 
 ```
--- a/server.js
+++ b/server.js
@@ -137,7 +137,7 @@ let botAudioManager = null;
 let socketRuntime = null;
 
 const UPLOAD_PATH_RE = /\/uploads\/((?!(?:bot-audio|deleted-attachments|stickers)\/)(?:[A-Za-z0-9_-]+\/)*[A-Za-z0-9_.-]+)/g;
-const UPLOAD_URL_PATH_RE = /\/uploads\/((?:[A-Za-z0-9_.~%-]+\/)*[A-Za-z0-9_.~%-]+)/g;
+const UPLOAD_URL_PATH_RE = /\/uploads\/+([-A-Za-z0-9_.~%/\\]+)/gi;
 
 function isSafeUploadRelPath(relPath) {
   if (typeof relPath !== 'string' || !relPath) return false;
@@ -172,9 +172,23 @@ function collectUploadRelPaths(contents) {
     UPLOAD_URL_PATH_RE.lastIndex = 0;
     let match;
     while ((match = UPLOAD_URL_PATH_RE.exec(content)) !== null) {
-      let relPath;
-      try { relPath = decodeURIComponent(match[1]); } catch { continue; }
-      if (/^(?:deleted-attachments|stickers)\//.test(relPath)) continue;
+      let decoded;
+      try { decoded = decodeURIComponent(match[1]); } catch { continue; }
+      const parts = [];
+      let escapesRoot = false;
+      const segments = decoded.split(process.platform === 'win32' ? /[\\/]+/ : /\/+/);
+      for (const segment of segments) {
+        if (!segment || segment === '.') continue;
+        if (segment === '..') {
+          if (parts.length === 0) { escapesRoot = true; break; }
+          parts.pop();
+        } else {
+          parts.push(segment);
+        }
+      }
+      if (escapesRoot) continue;
+      const relPath = parts.join('/');
+      if (/^(?:bot-audio|deleted-attachments|stickers)\//i.test(relPath)) continue;
       if (isSafeUploadRelPath(relPath)) paths.add(relPath);
     }
   }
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by ancsemi** [comment]
Ran it, all 60 tests pass including your integration test. The endpoint itself I have no complaints about: requireModBot, the limit validation, the recursive CTE, the cross-channel 409, and doing the whole thing in one transaction. FTS is maintained by triggers so the bulk delete cleans the search index for free.

The attachment side is where I would push back.

`relocateUnreferencedUploads` keeps a file when `latestDmMessage >= ownership.created_at`, and `latestDmMessage` is the newest DM anywhere on the server. Once anyone sends a single DM, every channel upload from before that moment satisfies the test, and keeps satisfying it. I checked on a scratch database: one DM and an upload recorded in the same second is already enough to retain it. On any server where people DM each other, that branch retains everything and no channel attachment is ever relocated.

The caution behind it is right. The server genuinely cannot read an encrypted DM to find out whether it references a file. But as written the check does not narrow that risk, it switches the cleanup off. Scoping it to DMs that could plausibly reference the file, or to whether E2E is in play at all, would leave something that still does its job.

That also lands on the existing single-message endpoint, since you rewired it through the same function. Today it relocates unconditionally; after this it would stop relocating in practice. That is a behaviour change to an endpoint this PR is not otherwise about, and I would rather see it on its own either way, with or without the DM rule.

Worth knowing about the scan too: the surviving-reference query is a `LIKE '%/uploads/%'` over the whole messages table, which cannot use an index. It exits early once the candidate set empties, and batching all the paths from a bulk delete into one pass is better than the per-path pattern already in the retention sweep, so this is not a complaint, just something to keep in mind if it ever runs somewhere with a large history.

One heads up for the rebase. server.js was missing `DELETED_ATTACHMENTS_DIR` from its import line, so every call into its copy of `moveUploadToDeleted` threw a ReferenceError. Two callers swallowed it and silently never relocated anything; the bot message delete did not, and returned a 500 with the row already gone from the database and no `message-deleted` broadcast, so the message stayed on everyone's screen. Your diff fixes it as a side effect of touching line 2. I have fixed it on main separately in 4e72e3a, so drop it from your branch.

**Comment by bernardokcosta** [comment]
> Ran it, all 60 tests pass including your integration test. The endpoint itself I have no complaints about: requireModBot, the limit validation, the recursive CTE, the cross-channel 409, and doing the whole thing in one transaction. FTS is maintained by triggers so the bulk delete cleans the search index for free.
> 
> The attachment side is where I would push back.
> 
> `relocateUnreferencedUploads` keeps a file when `latestDmMessage >= ownership.created_at`, and `latestDmMessage` is the newest DM anywhere on the server. Once anyone sends a single DM, every channel upload from before that moment satisfies the test, and keeps satisfying it. I checked on a scratch database: one DM and an upload recorded in the same second is already enough to retain it. On any server where people DM each other, that branch retains everything and no channel attachment is ever relocated.
> 
> The caution behind it is right. The server genuinely cannot read an encrypted DM to find out whether it references a file. But as written the check does not narrow that risk, it switches the cleanup off. Scoping it to DMs that could plausibly reference the file, or to whether E2E is in play at all, would leave something that still does its job.
> 
> That also lands on the existing single-message endpoint, since you rewired it through the same function. Today it relocates unconditionally; after this it would stop relocating in practice. That is a behaviour change to an endpoint this PR is not otherwise about, and I would rather see it on its own either way, with or without the DM rule.
> 
> Worth knowing about the scan too: the surviving-reference query is a `LIKE '%/uploads/%'` over the whole messages table, which cannot use an index. It exits early once the candidate set empties, and batching all the paths from a bulk delete into one pass is better than the per-path pattern already in the retention sweep, so this is not a complaint, just something to keep in mind if it ever runs somewhere with a large history.
> 
> One heads up for the rebase. server.js was missing `DELETED_ATTACHMENTS_DIR` from its import line, so every call into its copy of `moveUploadToDeleted` threw a ReferenceError. Two callers swallowed it and silently never relocated anything; the bot message delete did not, and returned a 500 with the row already gone from the database and no `message-deleted` broadcast, so the message stayed on everyone's screen. Your diff fixes it as a side effect of touching line 2. I have fixed it on main separately in [4e72e3a](https://github.com/ancsemi/Haven/commit/4e72e3a39063cd02b3f69fb316a7c6007f94c111), so drop it from your branch.

Rebased onto current main and addressed the attachment concerns. Encrypted-DM retention is now scoped to messages sent by the upload owner, so unrelated DMs no longer disable cleanup server-wide. The existing single-message webhook endpoint has been restored unchanged. I also expanded regression coverage for unrelated-user DMs, edited owner DMs, URL aliases/encoding, traversal rejection, protected references, and the original single-delete behavior. All 66 tests pass.

**Comment by ancsemi** [comment]
Both concerns are properly closed. I checked the scoping on a scratch database rather than reading it: an unrelated user's DM no longer retains the file, the owner's own DM still does, and the single-message endpoint relocates unconditionally again, verified live by posting an attachment through a bot and deleting it. Bulk delete works end to end, the limit rejects 0, 101, abc, -5 and 1.5, a bot without moderation gets 403 and a bogus token gets 404. All 66 tests pass.

I went to merge and hit a conflict, because #5540 landed a few minutes before this and the two touch the same line. It is one region:

```
const UPLOAD_PATH_RE = ...   <- main's version now excludes bot-audio
const UPLOAD_URL_PATH_RE = ...  <- yours, new
```

Keep main's `UPLOAD_PATH_RE` (the one with `bot-audio` in the negative lookahead) and add your `UPLOAD_URL_PATH_RE` line under it. Nothing else conflicts.

While you are in there, one thing worth folding in: `collectUploadRelPaths` skips `deleted-attachments|stickers` after decoding, and now that bot-audio is a protected directory it belongs in that list too. I do not think it is reachable today, since a bot audio URL is `/api/bot-audio/...` rather than `/uploads/bot-audio/...` and anything under there has no `upload_ownership` row to pass the channel-scope check, but the other three regexes all name it now and this one reading differently will confuse whoever touches it next.

Ping me once it is pushed and I will merge it.

bernardokcosta:
> Both concerns are properly closed. I checked the scoping on a scratch database rather than reading it: an unrelated user's DM no longer retains the file, the owner's own DM still does, and the single-message endpoint relocates unconditionally again, verified live by posting an attachment through a bot and deleting it. Bulk delete works end to end, the limit rejects 0, 101, abc, -5 and 1.5, a bot without moderation gets 403 and a bogus token gets 404. All 66 tests pass.
> 
> I went to merge and hit a conflict, because #5540 landed a few minutes before this and the two touch the same line. It is one region:
> 
> ```
> const UPLOAD_PATH_RE = ...   <- main's version now excludes bot-audio
> const UPLOAD_URL_PATH_RE = ...  <- yours, new
> ```
> 
> Keep main's `UPLOAD_PATH_RE` (the one with `bot-audio` in the negative lookahead) and add your `UPLOAD_URL_PATH_RE` line under it. Nothing else conflicts.
> 
> While you are in there, one thing worth folding in: `collectUploadRelPaths` skips `deleted-attachments|stickers` after decoding, and now that bot-audio is a protected directory it belongs in that list too. I do not think it is reachable today, since a bot audio URL is `/api/bot-audio/...` rather than `/uploads/bot-audio/...` and anything under there has no `upload_ownership` row to pass the channel-scope check, but the other three regexes all name it now and this one reading differently will confuse whoever touches it next.
> 
> Ping me once it is pushed and I will merge it.

Rebased onto the latest main and pushed. I kept main's UPLOAD_PATH_RE, retained UPLOAD_URL_PATH_RE below it, and added bot-audio to the decoded protected-directory skip in collectUploadRelPaths. The focused integration test and the full suite pass, 85/85. Ready to merge.

PR #5543, opened 2026-08-28
Fix dropped screen-share renegotiations

bernardokcosta:
## Summary

Fixes an intermittent issue where the screen sharer could see their local preview, while other participants saw a black screen or no stream tile. Leaving and rejoining the voice call temporarily restored the stream.

This affects the existing browser-based WebRTC screen-sharing path used by both web clients and Haven Desktop.

## Root cause

Two signaling problems could leave a peer permanently stuck:

1. The server silently rejected SDP offers and answers larger than 16 KB. Screen-share renegotiations can exceed this after ICE/STUN/TURN candidates have already been gathered.
2. When an answer was dropped or rejected, the sender remained indefinitely in `have-local-offer`. Further screen-share recovery requests were queued but could never run. Rejoining worked because it created a fresh `RTCPeerConnection`.

## Changes

- Increase the bounded SDP signaling limit from 16 KB to 64 KB.
- Add optional `offerId` correlation to WebRTC offers and answers.
- Ignore delayed answers belonging to an older offer.
- Roll back and retry unanswered offers with bounded backoff.
- Preserve recovery state if rollback fails.
- Stop automatic retries after the recovery budget is exhausted.
- Reset the retry budget when a new media or connection event requests negotiation.
- Clear pending offer timers during glare handling, peer removal, and channel rotation.
- Keep compatibility with clients that do not yet send `offerId`.
- Bump the `voice.js` cache version.

No native screen-capture transport or Haven Desktop-specific changes are required.

## Testing

- Added signaling tests covering:
  - SDP payloads above the previous 16 KB limit
  - offer/answer correlation
  - delayed and uncorrelated answers
  - unanswered-offer rollback and retry
  - rollback failure during timeout recovery and glare
  - bounded retry exhaustion and restart
  - `signalingstatechange` race guards
  - timer cleanup when removing a peer
- Full suite: `75/75` tests passing
- `node --check` passing
- `git diff --check` passing

### DIFF
--- a/public/app.html
+++ b/public/app.html
@@ -4130,7 +4130,7 @@
   <script src="/js/theme.js?v=4.0.1"></script>
   <script src="/js/notifications.js?v=3.16.15"></script>
   <script src="/js/servers.js?v=3.16.15"></script>
-  <script src="/js/voice.js?v=3.34.0"></script>
+  <script src="/js/voice.js?v=3.34.1"></script>
   <script src="/js/modmode.js?v=3.16.15"></script>
   <script src="/js/e2e.js?v=3.16.15"></script>
   <!-- Shared with the server (src/automod.js requires the same file) so both
--- a/public/js/voice.js
+++ b/public/js/voice.js
@@ -500,6 +500,7 @@ class VoiceManager {
 
       try {
         const conn = peer.connection;
+        peer._handlingRemoteOffer = true;
 
         // ── Glare handling (perfect negotiation) ──────────────
         //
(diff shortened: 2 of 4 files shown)


ancsemi:
Read it and ran it, 75/75 pass. The diagnosis matches what I can see in the code and the client-side work is careful: the offer-id correlation, the rollback-failure path rescheduling its own timeout instead of stranding the peer, `_handlingRemoteOffer` keeping the drain from firing mid-negotiation, and the timers actually being cleared on peer removal and channel rotation. The bounded retry with a reset on a fresh media event is the right shape, and the legacy-answer fallback means a mixed-version call still works.

One thing to change, and it is in the line the whole PR turns on.

`MAX_SDP_SIZE` is now 65536, and socket.io's `maxHttpBufferSize` is also exactly 65536. The frame carries the envelope too (event name, `code`, `targetUserId`, `offerId`), so an offer that passes your check can still be over the transport limit, and socket.io does not drop those, it kills the connection. I measured it against your branch:

```
offer JSON  65000 bytes -> fine
offer JSON  65400 bytes -> fine
offer JSON  65536 bytes -> DISCONNECTED (transport close)
```

So in the top ~130 bytes the failure mode goes from "silently rejected", which is the bug you are fixing, to "dropped out of the call", which is worse. A longer `offerId` shrinks that margin further, since you allow up to 96 characters.

I would lower `MAX_SDP_SIZE` rather than raise the buffer. `maxHttpBufferSize` was deliberately brought down from 1 MB and applies to every socket event, so raising it to make room for SDP gives that headroom to everything else too. Something like 48 KB keeps the app-level reject clean, is still triple the old limit, and should be well clear of the renegotiation sizes you were actually hitting. If you have measured real SDPs closer to 64 KB than that, say so and I will look at the buffer separately.

Everything else looks good to me. Push that and I will merge.

ancsemi:
Merged, and I made the size change myself rather than sending you round again for a constant.

`MAX_SDP_SIZE` is 49152 on main now, in 5ab16c8, with a comment explaining why it has to sit under `maxHttpBufferSize` rather than on it. Still triple the old limit and clear of the renegotiation sizes you were hitting. I checked afterwards that 49152, 49200 and 60000 byte offers are all refused cleanly with the connection intact, so an oversized SDP is an application-level no rather than a dropped call. Left the buffer alone for the reason above.

Everything else went in as you wrote it. The offer-id correlation and the recovery budget are the parts that make this hold up, and the tests are the useful kind. 83 tests passing on main now.

PR #5545, opened 2026-08-30
Add theme compatibility and recovery safeguards

bernardokcosta:
## Summary

- Add strict Theme API metadata parsing and compatibility classification.
- Preserve support for legacy themes without an API declaration.
- Block invalid and unsupported themes before CSS injection.
- Validate theme stylesheets again whenever they are served.
- Add extension Safe Mode with pre-paint theme and plugin suppression.
- Add recovery controls for resetting theme and plugin choices.
- Preserve preferences during transient `/api/themes` failures.
- Enforce compatible published and default themes on the server.
- Document Theme API compatibility and recovery behavior.

## Compatibility behavior

- `@haven-theme-api 1`: compatible.
- No declaration: legacy and allowed.
- Future positive API versions: unsupported and blocked.
- Empty, duplicate, malformed, or invalid declarations: blocked.
- Incompatible themes cannot be served, enabled, published, or selected as the server default.

## Recovery behavior

Safe Mode can be opened with:

`/app.html?haven-safe-mode=1`

While active:

- File themes and additive CSS themes are not injected.
- Plugin files are not fetched or evaluated.
- Haven uses its built-in base appearance.
- Installed extensions remain visible for diagnosis.
- Theme and plugin choices can be reset without removing unrelated preferences.
- Extensions remain suppressed until the reset is synchronized with the server.

Safe Mode and pending recovery fail closed even if the compatibility helper script is unavailable.

## Reliability and security

- Theme API responses are treated as authoritative only when the complete snapshot succeeds.
- Network and filesystem failures do not erase saved theme preferences.
- Removed themes are reconciled during refresh.
- Stale local fallback cannot overwrite a newer server preference.
- Alternate, nested, encoded, and hidden stylesheet paths are rejected.
- Existing invalid server defaults are ignored without silently rewriting stored data.

## Validation

- `node --test`: 120/120 passed.
- Locale validation passed.
- JavaScript syntax checks passed.
- `git diff --check` passed.
- Independent code review found no remaining actionable issues.


ancsemi:
Merged. Ran it merged against main before pulling the trigger, 120 of 120 pass and the locales still validate.

Safe Mode is the part I'm most glad to have. A theme or plugin that breaks the UI has been a dead end until now, with no way back in short of clearing storage by hand, and `?haven-safe-mode=1` failing closed even when the compat helper itself is missing is the right instinct for a recovery path.

The filename and request-path handling in themeMetadata got a proper look too, since it serves files off a directory from a user-supplied path. Encoded traversal lands on a 404 rather than a read, which is what I wanted to see.

PR #5548, opened 2026-09-01
Self Serve Roles (Groups)

birdcrazy:
Requested by Raidenphantom in the community server: 
  Asking for users to have the ability to add roles to themselves in order to gain channel access

### The Plan:
- Make use of the already existing role system.
- Allow level 0 roles (called "Groups") to be created.
- These roles don't grant any permissions, but can grant/revoke channel access.
- Users can join/leave these groups as they please with no elevated permissions needed.

### Implementation
- Allowed level 0 roles to be created and edited in the admin role manager and channel role manager.
  - Clearly indicate level 0 roles as "groups" and separate them a bit from actual leveled roles for ease of recognition.
  - Strip permissions from level 0 roles on role save
  - hide permissions from UI when editing level 0 role and display admin-note to describe why
- Added a "Groups" section to the user's profile settings modal.
  - This section only appears if level 0 roles exist on the server
  - Users can select which groups to join/leave on their own and they are assigned the roles to themselves.
    - Guards are in place server side to prevent a user from adding themselves to a leveled role.
  - Users can hover over a (#i) element for a popup that displays channels each "Group" grants them, so they can make informed decisions on which group(s) they want to join.

### Other Improvements/Fixes Included:
- Admin role manager now shows the level next to the roles in the role list (similar to the channel role manager)
- Bug fix, if the rename-modal was opened from the profile-popup before opening it from the user-bar, existing personas would not render
  - This was due to two separate blocks of code nearly identical running for each of these actions, and persona rendering was missing from one.
  - Fixed by commonizing the user-rename modal opening code into one function that both places now call.
- Audit log keys were but-to-but, with no space between them, added space between for more readability.
- Role editing always said everything was changed when it wasn't, fixed it to only report what actually changed.
- Added AI translations for all the newly introduced translation keys.

## Video Demo

https://github.com/user-attachments/assets/0919b4e0-85b2-4e43-8833-0b4e3f141294


ancsemi:
Merged, thanks. Two things I changed right after, both in 585a1b0.

Save Profile calls the group save on every save, and with the manager closed there are no checkboxes in the DOM, so it sent an empty list and the server read that as "leave every group". Changing your display name would have quietly dropped you from all of them. It now only sends while the manager is open, and the manager's checkboxes are cleared when it closes so an old selection can't be replayed later.

`get-role-channel-access` returned the whole channel table to anyone asking about a group, private channel names included. Non-role-managers now get just the channels the group grants.

Smaller stuff: put back `roles_created` in pl.json (it went missing in the diff), removed a stray Russian word from the zh note, and flattened the nested CSS rule since the stylesheet doesn't use nesting anywhere else. Ran the whole flow on a scratch server afterwards: join a group, get the channel, save the profile, still in the group.

PR #5549, opened 2026-09-01
Add Compact Theme API v1 theme

bernardokcosta:
## Summary

- ship an optional CSS-only Compact theme using Theme API v1
- add paired foreground tokens for success, danger, and warning controls
- migrate semantic and accent-filled controls to the appropriate foreground tokens
- document the new theme and public tokens
- add contract tests for selectors, responsive geometry, message density, semantic token pairing, and palette contrast

## Details

Compact uses a graphite and cyan palette with narrower desktop panels, tighter stable layout regions, reduced radii, and faster transitions.

The theme:

- uses only public Theme API v1 tokens, page markers, and layout regions
- applies compact geometry only above the 900px desktop breakpoint
- preserves the user's Layout Density setting
- keeps the existing tablet and mobile panel behavior
- does not include JavaScript or structural layout changes
- leaves structural compaction for a future Compact Layout plugin

The new `--success-text`, `--danger-text`, and `--warning-text` tokens preserve Haven's existing default appearance while allowing themes with bright semantic colors to provide readable foregrounds.

## Testing

- `npm test` — 123 tests passed
- focused theme suites — 38 tests passed
- `node scripts/validate-locales.js` — all 7 locale files passed
- `git diff --check`
- browser validation at 1440x900, 800x900, and 390x844
- login and application pages checked for horizontal overflow
- tablet and mobile off-canvas panels checked
- semantic normal and hover states checked through computed styles

### DIFF
--- a/GUIDE.md
+++ b/GUIDE.md
@@ -277,6 +277,11 @@ Click **Clear** to remove your avatar and revert to the default initial-letter a
 
 Haven includes 20+ visual themes. Click the **🎨** button at the bottom of the sidebar to open the theme picker. Themes change colors, fonts, and overall aesthetic. Your choice is saved per browser.
 
+Servers also include the optional **Compact** Theme API v1 file theme. An admin
+can publish it from **Settings → Admin → Branding → Custom Themes**. Compact
+tightens the desktop chrome while preserving the user's separate message-density
+choice under **Settings → Layout**.
+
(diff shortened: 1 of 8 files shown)


ancsemi:
Merged, thanks. The contract tests pass here too and the full suite is green on main with it in.

PR #5550, opened 2026-09-02
Images that show on Haven Mobile were blank on desktop

Amnibro:
## Why

Someone posted a photo (Ferry / bot, Discord CDN URL). Haven Mobile showed the cat. Desktop showed the text and nothing else — not even a broken-image icon. They had to ask "Can you see the photo?"

## What was going on

Ferry puts the attachment on its own line as a signed Discord URL (`?ex=…&is=…&hm=…`). Mobile loads that URL as-is. Desktop HTML-escapes the message *before* turning it into an ``, so `&` becomes `&amp;`. The media proxy then asks Discord for a URL it never signed, gets a 502, and the bubble looks empty.

Two more things on the same path:

- `_formatContent` was stricter than `_isImageUrl` for `/uploads/` filenames (no dots in the basename, no extra path segment), so a message classified as an image could emit no `` at all.
- The proxy sent `HavenBot` as the User-Agent (Discord 403s that) and rejected `application/octet-stream` even when the bytes were a JPEG/PNG.

## What changed

- Decode HTML entities before the media proxy / auto-link / markdown image handlers run.
- Remote image-only messages use the raw URL.
- `_isImageUrl` and the `/uploads/` render regex match each other (and mobile).
- Proxy uses a browser User-Agent, sends a Discord Referer for those hosts, and sniffs magic bytes when the CDN lies about Content-Type.

Needs a server restart (`mediaProxy.js`).

## Test plan
- [ ] Ferry / bot Discord photo with a signed query string: desktop shows the image, not a blank gap
- [ ] Same photo still shows on Haven Mobile
- [ ] `/uploads/photo.edit.jpg` and `/uploads/images/cat.webp` render as images
- [ ] `node --test test/chatImageUrls.test.js`

Made with [Cursor](https://cursor.com)


ancsemi:
Merged, thanks. Confirmed the escaped ampersand was the whole story on 4.2.0, and the magic-byte fallback in the proxy is a good catch in its own right. Goes out with the next release.

PR #5551, opened 2026-09-02
Internationalize the full client interface

bernardokcosta:
## Summary

- Internationalize user-facing UI across authentication, chat, media, voice, channels, administration, Ferry, desktop integrations, utilities, notifications, and built-in games.
- Expand the English and Portuguese catalogs while preserving English fallback behavior for partial locales.
- Add runtime support for additional translated DOM attributes, including labels, alt text, titles, placeholders, and ARIA labels.
- Translate dynamic messages, notifications, command output, administrative controls, role management, media galleries, and error states.
- Add automated coverage for catalog parity, placeholders, duplicate JSON keys, literal translation references, required HTML markup, language selection, and early password-control initialization.

## Testing

- `npm test` — 128 tests passed
- `node scripts/validate-locales.js` — all locale catalogs valid
- Static i18n audit — 0 missing references
- JavaScript syntax checks passed
- `git diff --check` passed
- Smoke-tested authentication in Portuguese on desktop and mobile
- Smoke-tested the authenticated application without console errors

Only `en.json` and `pt.json` are changed among the locale catalogs.


ancsemi:
Merged, thanks. This was a big one and it read cleanly; the suite is green at 131 here with it in.

One change after merging, in 36a71f3. The parity tests required pt.json to carry every English key and failed on any page reference Portuguese lacked. Haven's other locales are partial by design and fall back to English per key, and a contributor adding an English string usually cannot write the Portuguese for it, so the strict check would have failed every future PR that touches a string. What stays enforced: every referenced key exists in English, pt.json never carries a key English has dropped, shared keys keep matching placeholders, and the HTML-bearing strings keep their markup in every locale that has them.

PR #5554, opened 2026-09-03
[feat+QoL] Unified user context menu + profile card re-design 

Bo0sted:
## Overview

Reworks the user profile card and consolidates every user-action menu into a single right-click context menu. Removes the redundant hover card and the scattered gear-icon menus, and makes the card's avatar open full-resolution.

### Profile card
- **Removed the hover card.** The translucent hover-over preview is gone; hovering no longer pops anything up. The click card now uses that same translucent, theme-aware glass styling (was solid).
- **Anchored positioning.** The card opens to the right of whatever you clicked (avatar or name), vertically aligned with it, instead of centering above/below the cursor and covering content at high zoom. Falls back to the left side when it would overflow the right edge.
- **Click-to-toggle.** Clicking the same user whose card is open closes it; clicking a different user switches.
- **Clickable avatar.** Clicking a real avatar (not the letter fallback) opens the full-resolution image in the existing lightbox — for other users and yourself. The card stays open behind the lightbox, and dismissing the image keeps the card.
- **Removed** the card's "Set Nickname" and "More Actions" (gear) buttons — both now live in the context menu.

### Unified user context menu
Right-clicking a user  in the member list or directly on a message author's name/avatar in chat, opens one menu with everything, gated exactly as before.

- Section dividers only render when their section has visible items, so lower-permission users get a clean, shorter menu.
- Right-clicking your **own** message author falls through to the normal message context menu.
- The old mod-only gear menu (`_showUserGearMenu`) and all gear-icon entry points (profile card + member-list sidebar) are removed. `"Assign Role"` is relabeled **"Role Management"** across all 8 locales.

## Testing
- `node --test test/i18n.test.js` passes (12/12); all locale JSON validates.
- All modified frontend modules parse clean.
- Manually verified: hover card gone; card styling / positioning / toggle; avatar lightbox (opens over card, backdrop-dismiss keeps card); unified menu from both the member list and chat; permission gating per role; and the message menu still works for own / non-author right-clicks.

### DIFF
--- a/public/css/style.css
+++ b/public/css/style.css
@@ -6868,9 +6868,9 @@ details.sound-section:not([open]) .sound-section-label::before {
 .image-lightbox {
   position: fixed;
   inset: 0;
-  /* Sit above modal overlays (which use z-index: 100001) so opening a
-     lightbox from inside the media gallery actually appears in front. */
-  z-index: 100010;
+  /* Sit above modal overlays (100001) and the profile card (100010) so opening
+     a lightbox from the media gallery or a profile avatar appears in front. */
+  z-index: 100020;
   background: rgba(0, 0, 0, 0.85);
   display: flex;
   align-items: center;
--- a/public/css/style.css
+++ b/public/css/style.css
@@ -5835,22 +5835,6 @@ details.sound-section:not([open]) .sound-section-label::before {
 }
 .profile-edit-btn:hover { background: var(--bg-hover); }
 
-.profile-nick-btn {
-  background: var(--bg-tertiary);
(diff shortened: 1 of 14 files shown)


ancsemi:
Merged, thanks. Two adjustments went in right after in a follow-up commit: the hover preview card is back (there is now a toggle for it under Settings, Chat, on by default), and the gear buttons are back on member rows and the profile card. Instead of the old gear menu they open your unified right-click menu anchored to the button, so all the moderation actions live in one place with the divider you added. The rest is as you sent it, including the lightbox and the anchored card positioning.

PR #5555, opened 2026-09-03
Add negotiated native screen sharing transport

bernardokcosta:
## Summary

- Add a versioned native screen-sharing transport while preserving Chromium screen sharing as the default and fallback path.
- Negotiate screen-share capabilities across human voice participants, with H.264 as the required baseline and optional H.265/AV1 only when every participant advertises a common codec.
- Carry native screen audio and video in the same remote stream and keep screen state authoritative across reconnects, channel rotation, moderation, deletion, and disconnect flows.
- Add dedicated offer, answer, and ICE signaling with session and negotiation correlation, exact event allowlisting, bounded payloads, rate limits, and pending ICE limits.
- Reject incompatible senders and viewers before announcing a native share, and notify an active sharer when an incompatible client joins later.
- Add the desktop-only native transport opt-in and localized UI state.

## Compatibility

- Chromium capture remains unchanged for browsers and is still the default in Haven Desktop.
- Native capture is enabled locally with `haven_native_screen_share=1`.
- Voice bots are excluded from codec negotiation and do not block native sharing.
- Native signaling protocol version: `2`.

## Validation

- `npm test`: 234/234 tests passed after merging Haven 4.3.0.
- `git diff --check upstream/main...HEAD` passed.
- Final code review completed with no findings.

## Paired PR

Desktop capture and transport implementation: https://github.com/ancsemi/Haven-Desktop/pull/49

## Supersedes

This replaces #5535 with the complete capability negotiation, client transport, lifecycle, and signaling implementation.


ancsemi:
Merged, thanks. I pushed a small merge commit to your branch first to clear a conflict on the module cache-buster lines in app.js, nothing else changed. The server side is in and off by default, so it lands with the next release and only lights up once the Desktop half ships.

PR #5557, opened 2026-09-03
Ferry: relay a stream bot's link and thumbnail, not just its ping line

Amnibro:
## What broke

Stream and video promotion bots (CouchBot and friends) bridged through Ferry arrived in Haven as just the ping line. The Twitch / YouTube / Kick link and the thumbnail were gone.

Those bots type "@everyone X is now live" and put everything worth having in a **rich embed**: `embed.url` is the stream link, `embed.title` the stream title, `embed.image` / `embed.thumbnail` the preview picture. `buildHavenContent` only read embeds when the Discord message had **no** text, so the embed was skipped whenever the bot also typed a line, which is every promotion bot. And since 08e7417 an embed carrying an image drops `embed.url` on purpose, so even the text-less shape lost the stream link.

## What changed

`src/ferry.js` `buildHavenContent`:

- **Rich (bot-composed) embeds are read alongside the typed text.** Title, promoted link, and description join the message; the image / thumbnail goes into `media` and renders as an ordinary chat image (same path as Discord attachments, so it is not fed through the link filter).
- **The promoted link is policy-checked on its own.** On an allowlist server the link stands or falls by itself: twitch.tv and youtube.com are in the starter list so they come through, kick.com is not so that link is dropped while the announcement and its thumbnail still relay. Before, a blocked link would have taken the whole message down.
- **Discord's own unfurls are still ignored** when there is typed text (`type` of `link`, `video`, `article`, `image`, `gifv`), so Haven does not preview the same link twice. The link-only and SaucyBot shapes behave exactly as before.
- Title / link / description already present in the typed text are not repeated.

## Tests

- `test/ferry.test.js`: text + rich embed relays link, title, and thumbnail (Twitch image and YouTube thumbnail shapes); typed link is not duplicated; a typed link's Discord unfurl stays ignored.
- `test/ferryAutomod.test.js` (allowlist on): Twitch link kept; Kick link dropped, announcement and thumbnail kept.
- All three new tests fail on the unpatched `src/ferry.js`; 27/27 pass with the patch. The full suite shows the same 5 pre-existing failures on untouched `main` in my environment (audio inspection, forum channels, group E2E), unrelated to this change.

Rendering on the web client was already handled by #5550 (media proxy + raw URL decoding), so once Ferry hands the URLs over, desktop and mobile both show the picture.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,18 @@ Format follows [Keep a Changelog](https://keepachangelog.com/). Haven uses [Sema
 
 ## [Unreleased]
 
+### Fixed
+- **Stream and video announcement bots bridged through Ferry lost their link and
+  thumbnail.** CouchBot-style promotion bots type the ping line ("X is now live")
+  and put the stream link, title, and thumbnail in a rich embed. Ferry only read
+  the embed when the Discord message had no text, so Haven got the ping line and
+  nothing to click or look at. Rich (bot-composed) embeds are now read alongside
+  the typed text: the Twitch / YouTube / Kick link and the thumbnail come through
+  with it. The promoted link is checked against the link policy on its own, so an
+  allowlist server that has not added a host (kick.com is not in the starter list)
+  drops just that link and still relays the announcement and its picture. Discord's
+  own unfurl of a typed link is still ignored so Haven does not preview it twice.
+
 ### Added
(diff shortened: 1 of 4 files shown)


ancsemi:
Merged, thanks. The kick.com case dropping just the link and keeping the announcement is exactly right.

PR #5558, opened 2026-09-03
Admin Settings Visibility Refactor

birdcrazy:
Refactored `_syncSettingsNav()` to make admin settings access driven by a single dictionary.

The dictionary is used to determine which settings the user has access to; then the nav-items and settings sections visibility are set appropriately.

This makes the function much more maintainable and simple to adjust as more admin settings and roles are added/modified in the future.

A few missing nav-items were also added to round it out. (Includes AI translated keys for the added items)

---

Fixes the admin settings visibility issues from #5470


ancsemi:
Merged, thanks. One adjustment in a follow-up commit: the table had the Moderation section as admin only, which would have hidden it from moderators who hold ban, kick, audit log or view all members rights, so those are listed for it now, and the Admin tab gate in app.js carries the same three so it stays reachable for them. Having all of this in one table is a lot easier to keep straight than the old pile of special cases.

PR #5559, opened 2026-09-04
Add copy invite email card

birdcrazy:
Adds an additional copy button to invite links that copies a formatted html invitation card.

The user's currently active theme and language is applied to the card; As well as the server's custom branding.
The user can then paste this card into an email client to send to the invitee.

Upon merge or reject, #5470 can be closed.

image

image


ancsemi:
Merged, thanks. Closing #5470 with it, since the SMTP half is off the table and this covers the rest.

Issue #5564, opened 2026-09-04
Bug: Reply previews to bot messages show "[Deleted User]" as message author


delenda-delenda:
Reply previews ignore webhook authors (user_id is null), so they fall back to [Deleted User], rather than using the bot's name.

E.G Bot posts message:

> [BOT] GreetingBot
"Welcome to the server!"

When a user replies to that message:

> Replying to [BOT] GreetingBot: Welcome to the server!

shows above text input box but once posted, message shows:

> [Deleted User] Welcome to the server!

Haven version: v4.3.0

ancsemi:
Fixed by #5566, out with the next release. Thanks for the clear report.

PR #5566, opened 2026-09-04
Fix reply banners showing [Deleted User] for bot messages.

delenda-delenda:
Fixing https://github.com/ancsemi/Haven/issues/5564 

### Issue
Replying to a webhook bot message (e.g. LOLbot) showed `[Deleted User]` in the reply banner, while the original message correctly showed `[BOT] LOLbot `

Bot messages are stored with `user_id = NULL` and the display name on `webhook_username`. Reply enrichment only joined the users table, so the name fell through to the deleted-user fallback.

### Fix

- Shared `replyAuthorUsername` / `toReplyContext` in src/socketHandlers/helpers.js
- Reply enrichment (channel history, live send, threads) selects webhook/persona fields and formats authors like message rendering: `[BOT] ${webhook_username}`
- Webhook HTTP reply path in server.js uses the same helper

### Validation

- Unit: test/replyContext.test.js (bot / imported / persona / deleted / regular)
- Integration: test/replyToBot.integration.test.js (post as LOLbot → reply → assert live + history replyContext)
- Full suite: 192/192 passing (npm test)

Bug:
Screenshot 2026-09-04 at 18 43 51

After fix:
Screenshot 2026-09-04 at 18 54 25

### DIFF
--- a/server.js
+++ b/server.js
@@ -118,7 +118,7 @@ webpush.setVapidDetails(vapidEmail, process.env.VAPID_PUBLIC_KEY, process.env.VA
 
 const { initDatabase } = require('./src/database');
 const { router: authRoutes, authLimiter, verifyToken } = require('./src/auth');
-const { setupSocketHandlers, sanitizeText, sanitizeBorderTransform } = require('./src/socketHandlers');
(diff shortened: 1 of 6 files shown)


ancsemi:
Merged, thanks. Having the webhook HTTP path share the same helper means the three places can't drift apart again. It goes out with the next release.

Issue #5565, opened 2026-09-04
Feature Request: Support emoji/icons in soundboard


delenda-delenda:
Not sure if this is a feature request or a bug as emoji is shown in toast, but emojis don't seem to be supported in the soundboard.

Adding to soundboard sound name:
<img width="596" height="59" alt="Image" src="https://github.com/user-attachments/assets/ca4233c0-dafe-4499-bd6b-04a029782429" />

Works in success toast:
<img width="220" height="50" alt="Image" src="https://github.com/user-attachments/assets/f01b58e9-50cb-4df3-9d9a-2e54a2f499f3" />

However, emoji is not shown in the admin list or in soundboard panel:
<img width="584" height="37" alt="Image" src="https://github.com/user-attachments/assets/d79d73be-b764-4135-9a5d-1c584d2bb88c" />

ancsemi:
#5567 is merged, so emoji, flags and keycaps all survive now. Ships with the next release.

PR #5567, opened 2026-09-04
Allow emoji in soundboard names.

delenda-delenda:
Fixing https://github.com/ancsemi/Haven/issues/5565

### Issue
Soundboard custom names could include emoji when typing/saving, and the success toast showed them correctly, but the emoji never appeared in the soundboard or manage list afterward.

**Cause**: Upload (POST /api/upload-sound) and rename (PATCH /api/sounds/:name) sanitized names with /[^a-zA-Z0-9 _-]/g, which stripped all non-ASCII characters (including emoji) before writing to the DB. The toast used the client-side typed name; the listing reloaded from the API and showed the stripped value.

### Fix
Replaced the ASCII-only strip with sanitizeSoundName, which:

- Keeps letters/numbers (any script), spaces, _/-, and emoji (including ZWJ sequences / VS16)
- Strips noise like <> / punctuation
- Collapses whitespace and caps length at 30 code points
The helper lives in src/socketHandlers/helpers.js and is used by both upload and rename in server.js.

### Validation

- Manual: rename/upload a sound with emoji → toast and soundboard list both show the emoji after reload
- Automated: added test/sanitizeSoundName.test.js (emoji/ZWJ, stripping, whitespace, length, non-string)
- Full suite: npm test → 188/188 passed

image


ancsemi:
Merged, thanks. I added a small follow-up so flags and the # and * keycaps survive too, since the regional indicator symbols flags are built from aren't in Extended_Pictographic. Out with the next release.

PR #5570, opened 2026-09-05
[QoL] Remove Login EULA + Persist Modal closure to user-preferences

Bo0sted:
Two QoL improvements to streamline the login flow, especially for privacy-hardened browsers which clear localStorage every session.

## 1. Persist popup dismissals server-side

The desktop-app promo, mobile-app promo, and recovery-codes notice
remembered their dismissed state in localStorage, so browsers that wipe
storage re-showed all three on every login. Dismissal state now lives in
the existing per-account `user_preferences` table (`promo_seen_desktop`,
`promo_seen_android`, `recovery_notice_seen`) over the existing
get/set-preference channel. No schema change.

- Only the "don't show again" checkbox persists. A plain close is
  session-only and returns next login.
- The recovery notice also gates on server state: if the account already
  has recovery codes, it is skipped entirely.
- The top-bar app badges and every modal's call to action are unchanged.

## 2. Drop the consent gate on login

The 18+ and Terms checkboxes gated both registration and login, with
acceptance cached in localStorage, so hardened browsers forced a re-tick
on every login. Consent only needs capturing once, at signup, which the
server already records in `eula_acceptances` (version, IP, timestamp, age
flag). The checkboxes now appear only on the account-creation views
(register, SSO, guest), which still enforce them. Login drops the gate
and shows a note stating that logging in consents to the current Terms,
which may differ from the version accepted at registration, linked to the
Terms popup. Login still sends the version and age flag, so the server
contract is unchanged.

## Testing

Verified on a local instance: checkbox dismissals persist across reload
and a full localStorage wipe; plain closes do not; the recovery notice
respects both existing codes and dismissal; login works with no
checkboxes; registration, SSO, and guest still enforce them.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -37,6 +37,7 @@
           <input type="password" id="login-password" autocomplete="current-password" required>
         </div>
         <button type="submit" class="btn-primary" data-i18n="auth.login.submit">Login</button>
+        <p class="auth-consent-note" data-i18n-html="auth.login.consent_note" style="text-align:center;color:var(--text-muted);font-size:0.72rem;line-height:1.4;margin-top:10px">By clicking "Login", you formally consent to our <a href="#" id="login-tos-link">current ToS</a>, which may be different from the ToS that was effective when you first registered.</p>
         <div class="auth-recovery-links">
           <a href="#" id="forgot-password-show" data-i18n="auth.login.forgot_password">Forgot password?</a>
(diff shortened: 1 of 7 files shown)


ancsemi:
Merged, thanks. One follow-up on top: the old localStorage flags get carried over to the account once, so people who already closed the promos or the recovery notice on an older build don't see them again after upgrading. Your semantics for a plain close are kept as is.

PR #5575, opened 2026-09-05
Fetch cloudflared when it is missing, and say why Windows refused a port

Amnibro:
## What broke

A Windows self-hoster picked the **Cloudflare Tunnel** option in `Install Haven.ps1`, whose card reads "Free, no config needed. Auto-downloads cloudflared." Nothing downloads it. `src/tunnel.js` only ever ran `cloudflared` from PATH, so starting the tunnel from the admin panel failed with `cloudflared binary not found in PATH`, and the same person had already lost an afternoon to port forwarding.

## What changed

`src/tunnel.js`
- `resolveCloudflared()` looks on PATH first, then in `<data dir>/bin/cloudflared(.exe)`.
- `downloadCloudflared()` fetches the official release asset for the platform (`cloudflared-windows-amd64.exe`, `-386.exe`, `cloudflared-linux-{amd64,arm64,arm,386}`) from `github.com/cloudflare/cloudflared/releases/latest/download/` into `<data dir>/bin`, following GitHub's redirect, writing to a `.part` file and renaming, `chmod 755` on POSIX. macOS is not auto-downloaded (the release ships a tgz); the error says `brew install cloudflared`.
- `startTunnel('cloudflared')` uses the resolved path and downloads on first use. A failed download names the exact file to fetch and where to put it.
- `providerAvailable('cloudflared')` is true when the binary is present or downloadable, so the admin panel offers the option on a fresh box.

`Install Haven.ps1`: when the Cloudflare radio is chosen, pre-fetch the same file into `%APPDATA%\Haven\bin` (best effort, the server self-heals if it fails).

`server.js`: the `EACCES` bind message on Windows now points at `netsh interface ipv4 show excludedportrange protocol=tcp` and the `winnat` restart. On Windows that error is almost always a Hyper-V/WSL reserved range, and "ports below 1024 need elevation" sends people the wrong way.

## Verified

- `test/tunnelCloudflared.test.js`: asset names per platform, resolve order (PATH, then bin, then null), download into bin with rename and mode, failed download cleans up its `.part` and explains, availability reporting. 5/5.
- Live: with a temp `HAVEN_DATA_DIR`, `downloadCloudflared()` pulled 54,841,128 bytes through the GitHub redirect and the result ran: `cloudflared version 2026.8.3`.
- `Install Haven.ps1` parses (PowerShell AST parser, no errors). Existing `ferry` and `webhookCallback` suites still pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,20 @@ Format follows [Keep a Changelog](https://keepachangelog.com/). Haven uses [Sema
 
 ## [Unreleased]
 
+### Fixed
+- **Cloudflare Tunnel on Windows never worked out of the box.** The installer's Cloudflare
+  option says "auto-downloads cloudflared", but nothing downloaded it and the tunnel module
+  only ever ran `cloudflared` from PATH, so a fresh install that picked Cloudflare always
+  ended in "cloudflared binary not found in PATH" (reported by a self-hoster 2026-09-05).
+  The tunnel module now looks on PATH, then in `<data dir>/bin`, and fetches the official
+  release binary into that folder on first start when neither has it (Windows and Linux;
+  macOS gets a `brew install cloudflared` hint). A failed download says exactly which file
+  to drop where. The installer pre-fetches the same file when Cloudflare is chosen.
+- **A refused port on Windows now says why.** `EACCES` on Windows is almost always a port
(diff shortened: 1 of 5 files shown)


ancsemi:
Merged, thanks. That closes the gap between what the installer promised and what happened. I folded your changelog lines into the existing Fixed block so Unreleased has one of each heading, nothing else touched.

PR #5580, opened 2026-09-06
[feat] Timestamp generator modal

Bo0sted:
## Add a /time picker modal for inserting timestamps and a toolbar icon

Builds on the existing `<t:*:*>` timestamp rendering by adding a comprehensive timestamp generating modal.

### Entry points
- `/time` with no argument opens the picker. A typed argument still resolves directly and shows the usage toast on error.
- A clock button in the input toolbar, next to the poll button.

### The modal
- Date and time in labeled fields (Year/Month/Day, Hour/Minute/Second), with the native calendar picker reused from search and a 24HR/AM/PM toggle that defaults to the reader's own clock (via `Intl…hourCycle`, falling back to 24-hour).
- All seven styles are previewed live in the reader's locale, each with its own Insert button that drops the token at the caret, so it works mid-sentence.
- Invalid dates like Feb 31 disable insertion and show a placeholder.

### Notes
- Theme-aware and translation-driven. The usage hint reads straight from `commands.time_usage`, so locale edits flow through automatically.
- The clock button follows the other input controls and hides in media-only channels.
- The rendering pipeline (`_formatTimestampToken`) is reused unchanged. The modal only chooses the instant.

### DIFF
--- a/public/app.html
+++ b/public/app.html
@@ -554,6 +554,14 @@
                   <line x1="2"  y1="20" x2="22" y2="20"/>
                 </svg>
               </button>
+              <span class="input-actions-divider" id="time-divider"></span>
+              <button id="time-btn" class="btn-poll btn-time" data-i18n-title="app.input_bar.time_btn" title="Insert timestamp">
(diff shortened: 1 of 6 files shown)


ancsemi:
Merged, thanks. Reusing the render path meant nothing to re-review on that side. I bumped the cache busters in a follow-up commit. Goes out with the next release.

PR #5581, opened 2026-09-06
Change chat-image object-fit from "cover" to "contain"

birdcrazy:
Several people in the community server have complained about wide images cropping in the thumbnail image view.

This PR relieves this concern by changing the object-fit property to eliminate the cropping effect


ancsemi:
Merged, thanks. Cache buster bumped in a follow-up, out with the next release.

PR #5582, opened 2026-09-06
User Layout: Channel Scroll Behavior Settings

birdcrazy:
Adds a per-user channel scrolling setting that lets users choose between separate channel/DM scrollbars and a combined sidebar scrollbar.

Changes
- Added Channel Scrolling user setting with:
  - Separate — independent scrolling for channels and DMs.
  - Combined — one scrollbar for the channel/DM area. 
- Added a short-viewport fallback that forces the combined/full scrolling layout when vertical space is limited.
  - Extended the short-viewport behavior so the channel-management controls can scroll with the channel list, providing more usable channel space on smaller screens.
- Refactored the Layout settings section into separate Density and Channel Scrolling subsections.
- Added a shared picker setup helper to reduce duplicated picker initialization, persistence, validation, and active-state handling.


ancsemi:
Merged, thanks. One follow-up on my side: a source-level test that guards the data-haven-density hook was matching the old assignment form, so it now accepts the shared helper's key list. Cache busters bumped too. Next release.

PR #5583, opened 2026-09-06
Don't invite to all publics

### LINKED ISSUE #5569: Bug: Invite to every public channel when inviter doesn't have access to all public channels
If a user that is not part of every public channel creates an invite link. they only see the public channels they are in as options in the invite link creater/editor. 

Then if they don't uncheck any channels, the link defaults to granting access to "all public channels". 
Sure enough, when the invitee uses the invite, they are added to all public channels not just the ones the inviter sees on the invite creation.

I would expect only the channels the user can see selected on the invite creation should be the only channels that the invite grants access to.

Also, if all channel checkboxes are unchecked, the invite also links to all public channels

birdcrazy:
Fixes #5569

When an invite link had no channels selected to invite to, the invite would include all public channels, even though the user explicitly unchecked all channels.

The invitee would also get put into every public channel when the inviter created the link didn't have access to all public channels, and didn't uncheck any checkboxes.

This PR fixes this by only adding the invitee to the channels that were visible and explicitly selected when the invite link was created.

### DIFF
--- a/public/app.html
+++ b/public/app.html
@@ -3524,7 +3524,6 @@
           </label>
           <div>
             <h6 style="margin:0 0 4px;font-size:0.82rem;font-weight:600" data-i18n="settings.admin.invite_links.channels_granted">Channels this link grants</h6>
-            <small class="settings-hint" data-i18n="settings.admin.invite_links_channels_hint">Leave all checked to grant every public channel.</small>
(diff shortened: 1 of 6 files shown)


ancsemi:
Merged, thanks. One follow-up: links created before this stored an empty value for "every public channel", and the new code would have read that as "no channels", so those keep their old meaning while anything saved from now on is taken literally. Worth knowing for anyone reading later: a channel created after a link no longer gets added to it automatically. Closes out #5569.

PR #5584, opened 2026-09-07
Group Animated Image Settings with Image Display Settings

birdcrazy:
Similar to the setting groups introduced in #5582, this setting creates a new setting group in the "Image Display" section; Grouping `Image Display Mode`, `Animated Profile Pictures`, and `Animated Images in Chat` under the `🖼️ Image Display` settings section.

Relies on CSS style from #5582
(I didn't add the style in this PR to avoid conflicts)


ancsemi:
Merged, thanks. Landed after #5582 so the subsection styles were already in place. Next release.

PR #5585, opened 2026-09-07
Fix layout density buttons no longer working after #5582

birdcrazy:
#5582 broke the layout-density settings buttons.
This PR fixes the broken behavior by adding the necessary `buttonDataKey` to the `_setupPicker()` call.

Also fixed: when an array was assigned to `dataKey`, `buttonDataKey` used the value of `dataKey` by default, but didn't handle it as an array. It now defaults to using the the first element of `dataKeys` if `buttonDataKey` isn't provided.

I tested all the pickers after this change, and they are all now working.

### DIFF
--- a/public/js/modules/app-media.js
+++ b/public/js/modules/app-media.js
@@ -3692,7 +3692,7 @@ async _uploadBotAvatar(botId, file) {
 },
 
 // Helper function to simplify picker setups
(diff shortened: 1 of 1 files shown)


ancsemi:
Merged, thanks for the quick catch after #5582.

PR #5586, opened 2026-09-07
Generate the self-signed certificate in Node, no OpenSSL needed

Amnibro:
## What broke

MutantRabbit767 reported some devices never generate SSL certificates and then hit connection errors, even with OpenSSH installed. OpenSSH is not OpenSSL. Every Windows path (`Start Haven.bat`, `Install Haven.ps1`, the web installer) shelled out to `openssl.exe`, which Windows does not ship, so those machines silently got no certificate, ran plain HTTP, and every `https://` client failed.

## What changed

- `src/selfsignedCert.js` builds an RSA-2048 X.509 v3 certificate with Node's built-in `crypto` only. SAN covers `localhost`, `127.0.0.1`, every LAN IPv4 and the hostname. basicConstraints CA, serverAuth EKU, ten-year validity.
- `server.js` generates the pair at boot when the data directory has none and `FORCE_HTTP` is not set. Logs what it made.
- `Start Haven.bat`, `Install Haven.ps1` and `installer/server.js` call `scripts/gen-cert.js` instead of hunting for `openssl.exe`. The "Skipped (OpenSSL not found, will use HTTP)" outcome is gone.
- `FORCE_HTTP=true` still skips everything for reverse-proxy setups.
- `test/httpModeHeaders.test.js` now sets `FORCE_HTTP`, because that is the only way to reach plain HTTP now.
- GUIDE, support page and changelog updated. No new dependencies.

## Verification

- `test/selfsignedCert.test.js`: files written once, Node `X509Certificate` parses it, self-verify and private-key match pass, SAN checked, a real `https` server plus `https.get` with the cert as CA completes the handshake.
- External check: OpenSSL 3.5.5 `x509 -text`, `verify -CAfile`, `asn1parse` and `rsa -check` all OK on the output.
- Booted `server.js` against an empty `HAVEN_DATA_DIR`: cert generated, `HTTPS enabled`, `curl --cacert cert.pem https://localhost:PORT/api/health` returns 200.
- Ran the generator through `cmd.exe` and PowerShell the way the scripts do.
- The six tests that fail on this branch (bot audio, forumChannels, group-e2e-server, replyToBot) fail identically on current `main` in my environment.


ancsemi:
Merged, thanks. Main had picked up its own version of the generator while you were writing this one, so I merged yours over it: the scripts and installer now use the same generator and the certificate carries the CA and server auth flags, which is what phones want when you import it. The changelog entry moved under Unreleased and describes what is new since 4.5.0.

PR #5587, opened 2026-09-08
Load inline images on demand, closest first, and unload them far off screen

Amnibro:
## Why

Chat images, stickers and link-preview pictures fetch the moment a message renders, and all 100 messages kept in the DOM hold their decoded bitmaps. Measured on the Tauri desktop build against a busy channel, that was most of the renderer's memory: about 430 MB of a 700 MB process tree.

## What

A lazy media queue in `app-media.js`:

- The chat-image and sticker emitters in `app-utilities.js` hand the loader a blank placeholder plus `data-lazy-src` (`_lazySrcAttr`), so no fetch starts on render. A `MutationObserver` on `#messages` adopts link-preview images the same way.
- An `IntersectionObserver` with an 800 px margin marks what is near the viewport. Loads start closest first with at most three in flight: on-screen images immediately, the ones just below the fold in `requestIdleCallback` slices so a fast scroll never fights the fetches.
- A second observer with a 2400 px margin unloads what scrolled far away; a hidden document unloads everything after 20 s and reloads when it comes back.
- The rendered box is pinned on first load so unload and reload never move the chat.
- The lightbox reads the real URL through `_lazyRealSrc`.
- Cache busters bumped for app.js, app-media.js, app-ui.js, app-utilities.js and style.css.

## Verification

- `test/lazyMedia.test.js`: emitted markup, distance, pick order and parallel cap, lightbox URL.
- Real Chromium with an 80-image page wired to the module: 13 images fetched on open (viewport plus margin), 28 after jumping to the bottom, peak 3 in flight, images at the top unloaded, box pinned at its rendered size.


ancsemi:
Merged, with a fair bit added on top, so a heads up on what changed. The loader only watched #messages, so pictures in the thread panel, DM pop-outs and search results never loaded; it now watches the whole document. The observers were rooted on the viewport, which never sees anything scrolled out of a nested box, so the 800px margin did nothing inside the chat; each image is now observed relative to the box it scrolls in. Pinning pixel sizes fought the image size setting and window resizes, so an unloaded picture now gets a blank SVG of its own natural size instead. Frozen GIFs came back animated after a reload, so the freeze runs again. And when a picture above the reader took its size the chat slid by its height (Chrome's own anchoring did not catch it in that box), so the loader keeps its own anchor now. Checked in a real Chromium against a 40 picture history: main chat, thread, search, new picture while at the bottom, and mid history reading all hold still.

Issue #5578, opened 2026-09-06
[Feature] Updating plugins and themes from GitHub Releases


jkswoods:
Would you be interested in adding plugin/theme updates through GitHub Releases? I’m considering working on this, but wanted to get your thoughts before starting.

The idea is that plugins and themes include their repo in the existing metadata header:

```js
/**
 * @name Slate Layout
 * @id chat.username.slate-layout
 * @version 2.2.0
 * @update-repo username/haven-slate-layout
 * @update-channel stable
 */
```

Haven would periodically check that repo for newer compatible releases and show admins an “Update available” notice with the release notes. Clicking Update would download the `.plugin.js` or `.theme.css`, check it, back up the current file and replace it.

Since the files are shared across the instance, installation would be admin-only. Users would keep their enabled plugins and selected theme, with a prompt to reload once an update is installed. There should also be a way to roll back.

Each GitHub release could include a small `haven-release.json` alongside the files:

```json
{
  "schemaVersion": 1,
  "extensions": [
    {
      "id": "chat.username.slate-layout",
      "type": "plugin",
      "version": "2.2.0",
      "requires": {
        "haven": ">=4.3.0 <5.0.0"
      },
      "asset": "SlateLayout.plugin.js",
      "sha256": "<generated checksum>"
    }
  ]
}
```

The version range is just an example. A repo shipping both a plugin and a theme could list both in the same manifest.

It might also be useful to have an official reusable GitHub Actions workflow in the Haven repo. Authors could call it from their own repos to validate the metadata, generate the manifest/checksums and attach the files to their release. Releases would still live in each author’s repo; Haven would just provide the shared tooling.

I’d start with public repos and manually approved updates, plus a “Check for updates” button. No marketplace or automatic installs.

Does that sound like something you’d want in Haven? Open to suggestions on the metadata and how you’d prefer it to work.

And later on, an mvp official plugin/theme marketplace could just be a GitHub Pages site serving a directory of `haven-release.json` manifests. Haven could browse that directory and use the same install/update flow, with the actual files still hosted in each author’s GitHub Releases.

jkswoods:
... One other thought on security: since plugins run JS inside Haven, we’d want a few safeguards without turning this into a much bigger rewrite.

An install/update confirmation should explain what access plugins have and show the source repo and release notes, rather than just asking “are you sure?”. We could also require immutable GitHub Releases and verify the exact file checksum so assets can’t be swapped out after an update is offered.

[...]

PR #5588, opened 2026-09-08
Draft: plugin/theme release manifest and manual update proposal

jkswoods:
Related to #5578.

Adds manual, administrator-only updates for supported plugins and themes using public GitHub releases.

## How it works

Extension maintainers declare a stable ID, version, and public GitHub repository in the plugin or theme metadata:

```text
@id org.example.extension
@version 1.1.0
@update-repo owner/repository
```

Each immutable GitHub release contains the built `.plugin.js` or `.theme.css` file and a `haven-release.json` manifest describing its version, Haven compatibility, filename, and SHA-256 checksum.

Administrators can open Settings → Admin → Extension Updates and select Check for updates. Haven does not check or install updates automatically.

Before applying an update, Haven shows the repository, version change, release notes, and a security warning. The server then checks current administrator access, the security blocklist, release immutability, Haven compatibility, installed-file state, extension metadata, and the downloaded checksum before replacing the file.

The previous version is retained for rollback. Connected clients are notified when an extension changes and prompted to reload.

## Security

- Public GitHub repositories only
- Immutable, published, non-prerelease releases
- Required fresh blocklist check before updates or rollbacks
- SHA-256 verification before replacement
- Short-lived update offers bound to the administrator and installed file
- Administrator permission checked again immediately before replacement
- Download hosts, redirects, file sizes, and timeouts restricted
- Atomic file replacement with crash recovery
- Release notes rendered as escaped plain text
- Known-bad installed versions shown as warnings
- Known-bad updates and rollback versions blocked

If the blocklist cannot be fetched or validated, installed extensions continue running but updates and rollbacks are unavailable.

## Screenshots
Screenshot 2026-09-09 at 22 28 55
Screenshot 2026-09-09 at 21 27 43

## Documentation

The extension update documentation includes an administrator guide, extension maintainer instructions, metadata requirements, the `haven-release.json` format, blocklist behavior, and release checklist.

## Test plugin

A public test fixture is available at [jkswoods/haven-test-plugin](https://github.com/jkswoods/haven-test-plugin).

The repository contains version `1.0.0` for initial installation and an immutable [`v1.1.0` release](https://github.com/jkswoods/haven-test-plugin/releases/tag/v1.1.0) containing the updated `.plugin.js` file and matching `haven-release.json`. The downloaded release asset and manifest checksum have been verified.

The central `https://ancsemi.github.io/Haven/blocklist.json` currently returns `404`, so Haven correctly fails closed before offering the update. Once the blocklist is published, this plugin can be used to test update, reload notification, and rollback end to end.


ancsemi:
Went through the draft. The header and manifest shape fit how plugins and themes already declare themselves, so keep that as is. Yes to requiring immutable releases from the start, it is one setting on the publisher's side and it closes the swap-after-offer hole for good. And yes to failing closed: if the blocklist cannot be fetched, keep whatever is running but do not install or roll back until a fresh copy comes through, with a plain message saying why. Go ahead and build on that.

jkswoods:
@ancsemi Had a first pass at this, appreciate any feedback on the approach.

ancsemi:
Sorry this sat so long. I read through the updater and the admin endpoints, and the approach is what we talked about: nothing automatic, the admin rechecked right before the swap, immutable releases only, checksum before replacement, a backup for rollback, and failing closed on the blocklist. The release notes going through the escaping confirm dialog is good too.

A few things before it's ready:

1. One plugin with a broken header stops the whole check, since the error is thrown while reading the installed list. It would be nicer as an error on that one row, with the rest still checked.
2. The blocklist does not exist yet, which is why you get the 404. I'll publish an empty one at docs/blocklist.json when this goes in, so your test plugin can run end to end then.
3. main has moved a fair bit since the 9th (two releases), so it will need main merged in. It's usually just the ?v= lines that conflict.

Mark it ready when you're happy with it and I'll do a full pass and try it with your test plugin.

PR #5589, opened 2026-09-08
Visual effects survive a desktop relaunch (synced like the theme)

Amnibro:
Reported by Dispencer2 on the Windows desktop app: closing and reopening Haven Desktop resets the hand-picked visual effects to the theme defaults, while the theme itself stays.

**Root cause.** The effects pick lived only in `localStorage['haven_effects']`. The theme is mirrored to `user_preferences` and re-applied on every socket connect, so it is the only appearance setting that comes back when the desktop shell lands on a different storage origin (http vs https autodetect) or a wiped partition. With the key gone, `_getStoredEffectMode()` falls back to `auto`, which is the theme's default effects.

**Fix.**
- `src/socketHandlers/users.js`: `effects` is an allowed preference key; it carries a JSON array so its limit is 400 chars instead of 50.
- `public/js/theme.js`: every picker write goes through `_persistEffects`, which writes localStorage and emits `set-preference` like the theme does. `syncEffectsFromServer` restores the value and refreshes the picker buttons.
- `public/js/modules/app-socket.js`: the `preferences` handler restores effects before `applyThemeFromServer`, so the theme pass applies the saved pick in the same go.
- Cache busters bumped for theme.js and app-socket.js. Test in `test/effectsPreference.test.js`.

The same class of setting (`haven_fx_speeds`, CRT vignette/scanline, glitch frequency, zoom, density) is still localStorage-only; happy to follow up if you want those synced too.


ancsemi:
Merged, thanks. I added one small thing after it: with no personal theme and no server default theme there is no theme pass to carry the restored effects, so they sat in storage until a reload. The preferences handler now applies them itself in that case.

PR #5590, opened 2026-09-08
CRT theme text reads the same size as every other theme

Amnibro:
Reported by Dispencer2: the CRT style's font is inconsistent with all the others and far too small.

CRT is the only theme that swaps in VT323, a pixel face whose glyphs sit small in the em box. Measured in Chromium at 16px: cap height 10 vs Segoe UI's 11, and the same line of text 18 percent narrower, with thin single-pixel strokes on top. Nothing compensated for that.

Fix: `size-adjust: 120%` on the three VT323 `@font-face` blocks (latin, latin-ext, vietnamese all match, so mixed-script runs don't jump). That puts cap height and line width on Segoe UI's. `font-size` is unchanged, so padding, gaps and avatar sizes stay put. Supported in Chromium 92+ and Firefox 92+; older engines keep today's rendering. Cache buster bumped for style.css.


ancsemi:
Merged. size-adjust is a much lighter fix than special-casing the CRT font size everywhere, glad it exists.

PR #5591, opened 2026-09-08
Channel and DM lists auto-scroll while dragging near the edges

Amnibro:
Reported by Dispencer2 on the desktop app: holding a channel for drag-and-drop, the list won't scroll when the item reaches the top or bottom, so a long sidebar's off-screen channels are unreachable.

Chromium's native drag-and-drop only auto-scrolls the document, never a nested `overflow-y: auto` container, and the `dragover` handler in `_setupChannelDragDrop` only positioned the drop indicator.

**Fix.** `_makeEdgeScroller(listEl)` in `app-channels.js` drives the scroll from `dragover`: a 48 px hot zone at each end, speed scaling up to 18 px per frame at the very edge, a `requestAnimationFrame` loop so it keeps moving while the pointer holds still, and it resolves the element that actually scrolls on each call (`#channel-list` in separate mode, `.sidebar-split` in combined mode and under the short-screen media query). The channel list and the DM list both use it; `drop` and the cleanup path stop it.

Verified in a real Chromium on a 60-item list inside a 300 px scroller: scrolls down at the bottom edge, idles in the middle, scrolls back at the top edge, halts on stop. Test in `test/channelDragAutoscroll.test.js`. Cache buster bumped for app-channels.js.


ancsemi:
Merged, thanks. I merged main into your branch first to clear the version string conflict in app.js before squashing, nothing else touched.

PR #5592, opened 2026-09-08
Copy token button copies the token, or says it could not

Amnibro:
Reported by Dispencer2: the Copy button under Require invite token says "Token copied" but Ctrl+V pastes whatever was copied earlier.

The handler in `app-ui.js` called the success toast from the rejection path (`.catch(() => onCopied())`). In the desktop app `navigator.clipboard.writeText` is refused without a fresh user activation, so the toast lied while the clipboard kept its old contents. The neighbouring server-code button already has an `execCommand` fallback; this one had none.

**Fix.** The button tries the desktop bridge's main-process clipboard (`window.havenDesktop.clipboardWriteText`, not gesture-locked) first, then `navigator.clipboard.writeText`, then `_copyTextFallback`, and shows an error toast with a hint to select and Ctrl+C if all three fail. New string `settings.admin.registration.copy_failed` in en.json (other locales fall back per key). Cache buster bumped for app-ui.js.


ancsemi:
Merged. Same as the drag one, just a main merge on the branch for the version strings.

PR #5593, opened 2026-09-08
Soundboard hotkeys can be set and cleared from every layout

Amnibro:
Reported by Dispencer2: a soundboard hotkey can't be deleted from the list, you have to go back to the grid.

The sidebar list (`_renderSoundboardSidebar`) rendered a read-only hotkey chip with no Set link or clear control and bound only click-to-play. Hotkey management lived in `_renderSoundboard` alone. (The Sound Manager's own List view checkbox does show the ×, checked in Chromium, but it was a 13 px target.)

**Fix.** `_sbHotkeyControlsHtml` builds the chip-with-× or the Set link, `_bindSbHotkeyControls` attaches set, clear and right-click-to-record; the grid, the pop-out and the sidebar all use them. Row layouts get a finger-sized × and the pop-out grid carries `soundboard-grid` so its List view styles apply.

Verified in a real Chromium: clearing from the sidebar removes the binding, toasts and re-renders to the Set link; Set starts recording; clicks on the controls never play the sound. Test in `test/soundboardHotkeyControls.test.js`. Cache busters bumped for app-media.js and style.css.


ancsemi:
Merged, thanks. Small follow-up from me: giving the pop-out grid the soundboard-grid class also handed it that class's max-height and padding, so I zeroed those out on .sb-pip-grid so a resized pop-out fills its box.

Issue #5576, opened 2026-09-06
Settings Order Discrepancy


birdcrazy:
In both the user settings and admin settings, the order of the `settings-section`s  vs. `settings-nav-item`s don't match. 
Clicking down the `settings-nav-item`s one-by-one sends the `settings-body` scrolling up and down seemingly randomly.

birdcrazy:
I can work on this if you want. I just wanted to make sure there wasn't any particular reasoning for the order discrepancy. 

#5582 and #5584 should first get closed to avoid conflicts.

ancsemi:
Yes please, go ahead. #5582 and #5584 are both merged now, so the field is clear.

birdcrazy:
Done.
Submitted the PR (#5596).
Tested good on my end.

ancsemi:
Landed on main via #5596, thanks birdcrazy.

PR #5596, opened 2026-09-08
Reorder Settings Sections

birdcrazy:
From #5576:

Fixes the order discrepancy between the settings nav-items, and the settings sections that the nav-items point to.
Now the orders match for both user settings and admin settings.

Added some missing `settings-nav-item`s

Also fixed some minor naming and emoji inconsistencies. 

Added a scroll spy to admin settings just like the user settings already had. 
  Made a reusable scroll spy function that is used for both the user, and admin settings.
  
Updated translation keys where necessary.


ancsemi:
Merged, thanks. One small thing I fixed in a follow-up commit: the new Members heading had its label sitting outside the translated span, so it rendered as MembersMembers once the catalog filled the span in. Closing #5576 with this.

PR #5595, opened 2026-09-08
Forums read like forums: topic cards, gallery, sort, tags, New Post, plus NSFW channels, settings search, one create form

Amnibro:
Feedback from test2 (forum use on Discord: character showcases, RP posts, help forums with tags, list vs gallery, bump on reply) and Dispencer2's wishlist (NSFW toggle for phone use in public, search in Settings, merging the temp-channel button into the create form, & and + in channel names). Built on top of the bug-fix PRs from tonight; they merge independently.

## Server
- `messages.title`, `messages.tags` (JSON array of tag names), `channels.forum_tags` (JSON array of `{name, emoji}`), `channels.is_nsfw`. Migrations follow the existing patterns.
- `get-messages` on a forum accepts `sort: 'active' | 'created'`, `tags: [...]`, `tagMode: 'some' | 'all'`. Cursors resolve against the chosen key so paging stays consistent either way; tag filters use `json_each`.
- `send-message` stores a title and a pick of the channel's tags on forum topics only (unknown tags dropped, max 5). Payloads and history carry `title` and `tags`.
- `set-forum-tags` (channel managers, up to 20, deduplicated), `set-topic-meta` (author or manage_messages; emits `topic-updated`), `toggle-channel-permission` learns `nsfw`, `hide_nsfw` joins the allowed preferences. `channels-list` ships `forum_tags` and `is_nsfw`.
- Channel names may contain `&` and `+`.

## Web (`public/js/modules/app-forum.js`, a new mixin)
- Forum channels render as topic cards, newest activity on top: title (or first line), tags, author, reply count or "Reply to this topic", first image as thumbnail, pinned first.
- Gallery view: a grid of true squares, image fills the tile, text on a scrim. List/gallery, sort and tag filter persist per channel.
- Toolbar: New Post (title, body, tag picker), sort, view toggle, tag chips with match some/all.
- Replies bump the topic and refresh its count, new topics land on top, "Load more topics" pages older ones, `topic-updated` and `forum-tags-updated` refresh live. Edit title and tags from the card (author or moderator). Admins edit the tag list from Channel Functions.
- Channel Functions: NSFW switch (🔞 in the sidebar); Settings: Hide NSFW channels (keeps the one you're in visible).
- Settings search: a box at the top of the nav, Ctrl+F while Settings is open, hides sections and nav entries that don't mention the word.
- Create Temp Channel opens the create form with Temporary ticked.
- Cache busters bumped for every touched module.

## Verification
[...]

PR #5597, opened 2026-09-08
Role templates, required roles (any/all), per-role upload caps, channel templates, self-assign role menus

Amnibro:
The rest of Dispencer2's wishlist, on top of the bug-fix PRs (#5589-#5593) and the forum work (#5595). Each one is independent; happy to split if you'd rather take them one at a time.

## Role templates
New Role in Role Management opens a picker: Moderator, Helper, Trusted member, Event host, Media poster, Group, Blank. Each comes with a level, colour and permission set already ticked; name and level are still asked for, and the role is fully editable afterwards. Non-admins still only get the permissions they hold (server-side filter unchanged).

## Required roles on a channel, any or all
Channel Functions gets a **Required roles** row. Pick roles and choose **Any** or **All**. The gate sits on top of membership:
- `channels.role_gate` JSON `{mode, roles}`; `getEnrichedChannels`, `get-messages` and `send-message` check it.
- Losing a role leaves the socket room live and pushes a fresh list; gaining one opens it.
- Admins are exempt, like private channels. `delete-role` strips the role from every gate.
- New event `set-channel-role-gate` (manage_channel_settings), broadcast `channel-role-gate-updated`.

## Per-role upload cap
`roles.max_upload_mb` (blank = server default). The server-wide Max Upload Size is the floor; a role raises it for its holders; the highest cap among a user's roles wins. `/api/upload` and `/api/upload-file` check the effective cap, the disk-guard margin uses the largest cap in play, and `get-server-settings` reports `max_upload_mb_effective` per user so the composer's pre-check matches. A role change pushes the new number.

## Channel templates
Create Channel has a template picker: Chat, Announcements (read-only + announcement mode + everyone added), Forum, Private team, Event (temporary 24 h), Slow chat, Text only. Any channel can be saved as a template from Channel Functions (**Save as template**, manage_server) into the `channel_templates` setting. `create-channel` now accepts `topic`, `readOnly`, `announcement`, `slowMode`, `mediaEnabled`, `voiceEnabled`.

## Self-assign role menus
Settings, Roles, **Post a role menu**: pick a channel, a title, and roles with an emoji each. The message shows a button per role; clicking it, or reacting with the emoji, grants the role, and undoing either drops it. Reaction chips are seeded so they are one click away. Stored in a `role_menus` table keyed by message id; every reader receives their own `held` list.

## Tests
`test/rolesWishlist.test.js` boots a server on port 3397 and covers: cap on the role, ALL gate hides until both roles held, ANY gate, delete-role strips the gate, effective cap (2 MB upload accepted with a 5 MB role cap against a 1 MB server cap, refused without), role menu by click and by reaction, denial for non-managers, and template fields on create-channel. Full suite green apart from an unrelated port flake in `replyToBot.integration` that passes on its own.

Cache busters bumped for every touched module plus the new `app-role-tools.js`.


ancsemi:
Merged, thanks. Your branch was from before 4.6.0, so I merged main into it and resolved the overlap (the channel list columns, the reorganised settings page, cache busters) before squashing; your test file passes on the merged tree. Two small things I tightened in a follow-up commit: the gate now also covers thread reads, thread replies and voice joins, which only checked membership, and the upload cap counts server-wide roles only, the same as the client's held list. All of it goes out with the next release.

Issue #5600, opened 2026-09-09
Bug: UI Display Issues


TianLaiEric:
<img width="1098" height="926" alt="Image" src="https://github.com/user-attachments/assets/adde3613-ff17-4c8d-b80f-5282265b8d7b" />

1. The image sometimes appears and sometimes does not.
2. The size of the button for sending private messages is not symmetrical / properly aligned.
3. Feature suggestion: Please consider adding an attachment upload button to private messages, so users can send files or images directly.

birdcrazy:
1. What haven version are you on? (2nd report of #5574)
2. Good feedback, will look into it.
3. If you expand the PiP to full screen, the attachment options are present.

TianLaiEric:
Haven Server: 4.6.0
Haven Desktop: 1.4.30

ancsemi:
Number 1 is the same thing as #5574 and is fixed now: the PiP only looked at the online list for the channel you were viewing, so a DM partner outside that channel showed as an initial with a grey dot. Number 2: the send button now has the same height as the reply box, so it lines up in every theme. Number 3: as birdcrazy said, expanding the PiP gives you the attachment buttons. The two fixes go out with the next release.

PR #5601, opened 2026-09-09
Style dm-pip-send button to match thread-send-btn

birdcrazy:
Reported in #5600

> The size of the button for sending private messages is not symmetrical / properly aligned.

This PR solves it by making the style applied to `dm-pip-send` button the same as the style on the `thread-send-btn`


ancsemi:
Merged, thanks. I had already landed a smaller height-only tweak for #5600 before I saw this, so I merged main into your branch and kept your version of the rule, which is the better one. Goes out with the next release.

PR #5621, opened 2026-09-09
Add underlines

birdcrazy:
Adds `__underlined text__` 
and `Ctrl/Cmd + U` Keyboard shortcut to apply underlines wrapper around selected text.


ancsemi:
Merged, thanks. Nice and small. Goes out with the next release.

PR #5623, opened 2026-09-09
Add Inline Message Text Colors

birdcrazy:
Adds a formatting option to render custom colors within the message body.
Two color formats are supported:
- `c#(R,G,B)colored text#c` where R, G, and B are the red, green and blue values respectively in the range of 0-255.
- `c#RRGGBBcolored text#c` where RR, GG, BB are the red, green and blue values respectively in hexadecimal notation.

No dedicated picker modal yet, maybe in the future.. for now colors can be found through sites like [w3schools](https://www.w3schools.com/colors/colors_picker.asp).

Keyboard shortcut `Ctrl/Cmd + Shift + F` (F for "font", since c for code is already taken) inserts the `c#(R,G,B)...#c` formatting around selected text and moves the curser inside the parentheses for convenience to change the color quickly.

I'm open to changing the `c#...#c` formatting wrapper if you think something else is more appropriate. 
I arrived at this wrapper because I think discords color formatting is impossible, requiring external tools to use.. and used "c" for color, and # for hex color codes.

### DIFF
--- a/public/js/modules/app-admin.js
+++ b/public/js/modules/app-admin.js
@@ -2480,8 +2480,8 @@ _handleMarkdownShortcuts(inputEl, event) {
     return this._wrapSelectedText(input, '||', `||`);
   }
 
-  // Text color (Ctrl/Cmd + Shift + C)
-  if (event.shiftKey && key === 'c') {
+  // Text color (Ctrl/Cmd + Shift + F) (F for font, since c for code is already taken)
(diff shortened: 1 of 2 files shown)


ancsemi:
Merged, thanks. The c# wrapper reads fine to me and it is a lot friendlier than the ANSI trick Discord needs. Goes out with the next release.

PR #5628, opened 2026-09-10
Stop Braid and Compact file themes from stacking on Matrix

Amnibro:
## Summary
- File themes that set `--bg-primary` (Braid, Braid Light, Compact) are exclusive palettes. Turning them on in Settings selects that palette instead of stacking their `:root` tokens on Matrix.
- Theme → Layout has Braid and Compact switches so chrome stays independent of the picker colors.
- Picking a built-in theme persists first, then reapplies only real CSS tweaks, so Settings toggles match the picker.

## Test plan
- [ ] In Settings → Plugins & Themes, turn on Braid / Braid Light / Compact theme toggles, then pick Matrix. Matrix colors should apply everywhere.
- [ ] Theme → Layout: Braid on + Matrix still Matrix colors with Braid chrome.
- [ ] Theme → Layout: Compact on + Matrix still Matrix colors with Compact chrome.
- [ ] Picking Braid or Compact as the theme still applies that palette everywhere, including under the matching layout.
- [ ] Unpublished tweak themes that do not set `--bg-primary` still stack on a built-in theme.

Made with [Cursor](https://cursor.com)


ancsemi:
Merged, thanks. Splitting layout from palette means the Settings toggles and the picker finally agree with each other.

PR #5629, opened 2026-09-10
[feat] Add KLIPY as GIF provider

Bo0sted:
## Add KLIPY as GIF provider

Adds KLIPY as a third GIF provider alongside GIPHY and Tenor.

Two new variables:
- `KLIPY_API_KEY` — enable KLIPY
- `PREFERRED_GIF_SEARCH` — `klipy`, `giphy`, or `tenor`; picks which provider serves the picker when more than one key is set

If `PREFERRED_GIF_SEARCH` is unset or invalid, it falls back to whatever is configured: GIPHY first, then KLIPY, then Tenor last (Tenor is deprecated). This replaces the old rule where a GIPHY key simply took precedence over Tenor.

Supplemental and backwards-compatible — existing GIPHY/Tenor setups are unchanged. KLIPY results normalize to the same shape the client already expects, and the picker footer attributes KLIPY when it serves.


ancsemi:
Merged, thank you! Good to have a third option with Tenor on the way out. The fallback order and the preferred setting both read cleanly.

PR #5634, opened 2026-09-10
Add theme aware style to new-channel-template-row

birdcrazy:
The new `new-channel-template-row` dropdown didn't have any theme aware styling applied and the white background color protruded the boarder.

This simply adds the same class to the element that the settings dropdowns have, allowing the dropdown to be styled similarly.

---

Before:
image

After:
image


ancsemi:
Merged, thanks. The template dropdown matches the rest of the theme now.

PR #5667, opened 2026-09-14
[feat] Persisted timezone and format

Bo0sted:
## Persisted timezone and time format

Times in Haven followed whatever the browser reported for the zone and the
12/24 hour convention. That breaks in two ways: hardened/privacy browsers
deliberately report a false zone to resist fingerprinting, and ordinary
browsers report the zone and clock format from region settings that are often
wrong or not what the user wants. This PR lets a user set their timezone and
clock format once, persisted to their account and applied on every device,
independent of what the browser claims.

### Storage
Reuses the existing `user_preferences` table with three keys: `timezone` (an
IANA zone id such as `America/New_York`), `time_format` (`12` or `24`), and
`tz_prompt`. Zones are stored as IANA ids rather than fixed offsets, so `Intl`
resolves the correct offset per instant and daylight saving is always applied.

### Settings and first-run prompt
The Language settings section is renamed Localization and gains a Configure
Time button. It opens a modal to pick a zone from the full IANA list
(defaulting to the browser's current zone), choose 12 or 24 hour time, and see
a live preview. On the first login after the update the modal offers Skip
(never ask again), Remind later (ask again next launch), or Confirm. Guests are
never prompted.

### Clear button
Once a timezone has been configured and a user re-opens the modal from settings,
the "Skip" button is hidden and a new "Erase" button is presented. "Erase" does
exactly as it claims, erasing the user's configured timezone and allows Haven
to fallback to the default behavior for determining time based on what their
browser reports. 

### Where the setting applies
- **Display:** all timestamp rendering routes through one set of formatters, so
  chat times and tooltips, `<t:>` tokens, edited tags, forum posts, profile
  join dates, admin ban/audit/session lists, invite expiry, and the status bar
  clock all follow the setting.
- **Timestamp creation:** the `<t:>` generator modal and the `/time` command
  read the entered wall-clock in the configured zone.
- **Send later:** now uses the same custom wall-clock picker as `/time` instead
  of a native `datetime-local`, so a browser reporting a false zone no longer
  schedules the message at the wrong real-world time. The server handoff is
  unchanged; it still receives a plain absolute UTC instant.

### Fallback
With no timezone confirmed, every path falls back to the browser exactly as
before. Skip and Remind later change nothing about how times are shown or
interpreted.


Bo0sted:
Fixing merge conflict now

ancsemi:
Merged, thanks. Two tweaks on top of it on main: the modal no longer opens by itself at login, it lives under Settings, Localization, Configure Time (Skip and Remind later went with that, Cancel sits next to Erase and Confirm now), and a saved zone the browser does not recognise falls back to the browser's own zone instead of breaking the message list. Next release.

PR #5668, opened 2026-09-15
[fix] Edit message broke Ctrl+E emoji menu

Bo0sted:
The edit textarea's keydown handler stopPropagation's every key, so the global Ctrl+E shortcut never fired during an edit and the picker wouldn't open. This re-handles Ctrl/Cmd+E locally in the edit textarea, reusing the existing `_activeEditTextarea` routing so the picked emoji lands in the edit box.


ancsemi:
Merged, thanks.

PR #5669, opened 2026-09-15
Hide Banner nav-item when no banner is present

birdcrazy:
Simply hides the banner nav-item (Just as the banner settings section is already hidden) when no banner is defined in admin settings.


ancsemi:
Merged, thanks.

PR #5673, opened 2026-09-16
Add Haven Glyphs interface icon plugin

bernardokcosta:
## Summary
- Add the optional Haven Glyphs plugin for local Font Awesome Solid interface glyphs.
- Preserve message bodies, reactions, emoji/GIF pickers, soundboard names, user content, activities, and form fields.
- Keep icon conversion reversible and scoped to known UI hosts, with leading-only conversion for mixed labels.
- Bundle the font and third-party license notice.

## Validation
- `node --test --test-concurrency=1 test/havenGlyphs.test.js test/themeContract.test.js`
- `node --check plugins/HavenGlyphs.plugin.js`
- Installed in the Docker container on `10.84.22.173`; `/api/plugins` reports Haven Glyphs, the font is present, and the container is healthy.

The Hacord theme/layout is intentionally not included in this PR.


ancsemi:
Merged, thanks. I checked the font against the official 6.7.2 file, byte for byte. Two things on top of it on main: Settings, Interface Icons (the old Toolbar Icons) now has Glyphs as a third choice next to Monochrome and Colorful Emoji, and picking it turns the plugin on, so nobody has to find it on the plugin page; and the channel template names keep their emoji, since those sit in a dropdown the plugin cannot reach and everyone without the plugin was losing them.

PR #5658, opened 2026-09-13
Permissions matrix, original layout, and image fixes

Amnibro:
## Summary
- Settings → Permissions is a grid of role columns and a Users tab. New servers start with Member and Mod. Admin stays `is_admin`.
- Save Image uses a folder picker where the browser has one. Chat pictures that painted black in some Tauri profiles load again. Chrome's local-network prompt waits until someone joins voice.
- Theme → Layout has Haven (original). Sidebar chrome (DMs button, home-icon menu, dock chips) is on this branch, merged onto 4.8.0.

Role templates already landed in #5597. This is the leftover client pile on current main.

## Test plan
- [ ] Host opens Settings → Permissions and can add a role, tick perms, and assign someone on the Users tab
- [ ] Reset to Default seeds Member + Mod
- [ ] Save Image in Chrome opens a picker; Cancel shows a toast
- [ ] Opening chat does not prompt for local-network access; joining voice still can
- [ ] A chat PNG in Tauri is the picture, not a black box
- [ ] Theme → Layout original / Braid / Compact stay exclusive
- [ ] Forum still has feed, unread dots, shape picker, Mark all read, and Set as default
- [ ] 4.8.0 required-roles membership, send later, and poll pictures still work

Made with [Cursor](https://cursor.com)

PR #5675, opened 2026-09-17
Permissions grid with Users tab and Member/Mod defaults

Amnibro:
## Summary
- Settings → Permissions is a grid of role columns plus a Users tab
- New servers and Reset to Default seed Member (auto-assign) and Mod; the host stays Admin via `is_admin`
- More options still opens the existing role editor

Split out of #5658, first piece as requested.

## Test plan
- [ ] Open Settings → Permissions and add a role, tick perms, assign someone on the Users tab
- [ ] Reset to Default seeds Member + Mod
- [ ] Host still has every permission without a role row
- [ ] More options still opens the classic role editor


ancsemi:
Went through this one line by line since it is the permissions system. The grid itself is good and I want it in. A few things need fixing first:

- Picking a role on the Users tab removes every other server role the person has. People can hold several roles here, and the role menus from 4.8 hand them out, so one click in the grid would wipe those. It should add or remove the one role, the way assign-role and revoke-role do. Those two already have the level checks and the audit log entries, so calling them from the chips would cover both.
- Ticking a permission for one person sends only the boxes the grid shows, and the server treats anything missing as denied. So a permission that is not in the grid (transfer_admin today, manage_tags from the tagging PR that just merged) gets switched off for that person as a side effect, and when the editor is not the admin, so does every permission the editor does not hold. assign-role has a block for exactly this (the one that inherits overrides the caller may not touch) that is worth copying.
- The new handlers do not write to the audit log, and set-user-server-role does not look at the level of the person being changed, so someone with manage_roles could demote a person ranked above them.
- The Member role template has no label in en.json, so the template picker shows the raw keys (settings.admin.role_templates.member and member_desc).
- Dropping Channel Mod from the defaults switches off the channel creator role on new servers, since the default there is the highest channel role. I would keep Channel Mod in the seed.

I am still thinking about the wider Mod defaults (ban, IP ban, promote, audit log), so leave those as they are for now. It will also need a rebase: the tagging PR touched the same seed list in database.js.

Amnibro:
Rebased on main and went through your list.

- A chip on the Users tab now adds or removes that one role, through assign-role and revoke-role, so the other roles a person holds stay. The replace-everything handler is gone.
- A tick for one person sends the permissions the grid shows, and the server keeps every other one as it is (transfer_admin, manage_tags, anything added later). For an editor who is not the admin it also inherits every permission they may not touch, same as the block in assign-role.
- set-user-server-perms refuses a target at or above the editor's level and writes a `user_perms_update` audit entry with what was granted and denied.
- Member template has its label and description in en.json.
- Channel Mod is back in the seed, and Mod gets Manage Tags the way the Server Mod on main does. The wider Mod defaults are as they were.
- Manage Tags shows in the grid.

test/permGrid.test.js boots a server and covers the seed, the chips, the unseen-permission case, the non-admin editor limits, and Reset to Default. Full suite passes.

ancsemi:
Merged, and out in 4.10.0. Thanks for turning the list around that fast, and for the test file; I ran it and clicked through the grid and the Users tab on a test server before merging. The wider Mod defaults are fine as they are, so nothing more to do there.

PR #5676, opened 2026-09-17
Save Image picker and Tauri black picture fix

Amnibro:
## Summary
- Save Image uses the browser folder picker where one exists; Cancel is a toast
- E2E chat pictures that went black in some Tauri webviews keep their blob URL for a minute

Split out of #5658. Ferry gallery and the STUN defer are already on main, so they are not in this PR.

## Test plan
- [ ] Save Image in Chrome opens a picker; Cancel shows a toast
- [ ] A chat PNG in Tauri is the picture, not a black box
- [ ] Desktop `havenDesktop.saveImage` still works when present


ancsemi:
Merged, after bringing the branch up to main since the changelog line clashed with 5677. One change on top: Save was picking up the copy that Copy Image warms, which is a PNG re-encode, so a GIF saved as one still frame under a .gif name. It fetches the file itself now and keeps the PNG copy as a fallback only, decrypts an encrypted DM picture again the way Open does, and looks up a lazy-loaded picture's real address first.

PR #5677, opened 2026-09-17
Channel Functions: turn off reactions

Amnibro:
## Summary
- Channel Functions has a Reactions ON/OFF switch (default on)
- Off hides the react control and the server refuses add/remove; existing badges stay
- DMs stay on

Asked for by Andalishious in that same Channel Functions menu.

## Test plan
- [ ] Admin turns Reactions off; the smile button is gone and clicking an old badge does nothing
- [ ] Turning it back on restores reacting
- [ ] DMs still accept reactions


ancsemi:
Merged. Checked it on a test server: off hides the button and the server refuses the reaction, back on restores it.

Issue #5599, opened 2026-09-09
Tags on images in art channels, so the media gallery can be searched


ancsemi:
From Dispencer2 on the community server.

Art channels pile up. With a few artists posting a couple of times a week, a year in it gets hard to scroll back and find that one image someone posted about that one design. The Files & Media gallery already has a search, so the ask is: let people put a few tags on an image in specific channels (art channels, mostly), and have the gallery search and filter on them.

Rough shape:
- A tag field on an image in the gallery, and maybe at upload time, in channels where an admin turns it on.
- Gallery search matches tags as well as file names, plus a tag filter like the one forum channels got in 4.6.0.
- Tags per image rather than per message, since one message can carry several pictures.

Not urgent, a quality of life thing.

quakeman00:
i am dispencer2, lol, but thank you for keeping track of it

ancsemi:
This is in on main, built by Bo0sted in #5678. Tags go on a picture or file as you send it, or later from the message menu, Files & Media has a Tags filter, and search takes tag:. Making a new tag needs a Manage Tags permission; everyone can use the ones that exist. It works in every channel except DMs and forums rather than being switched on per channel. Next release.

PR #5678, opened 2026-09-18
[big feat] Attachment tagging

Bo0sted:
### Summary
Adds an optional, global tag system for file and image uploads: attach tags when you send an attachment, edit them later, search by them, and manage the vocabulary as an admin. Tags are a single server wide vocabulary applied to attachments, kept fully separate from the existing per channel forum topic tags. Ships across the composer, message display, search, the retroactive editor, and the media gallery, with admin configurable limits and a management panel.

### Background
Haven has no attachments table: an upload is a `messages` row whose content carries the `/uploads/...` URL markdown. Upload tags therefore key on `(message_id, rel_path)`. This feature is distinct from forum topic tags (`channels.forum_tags` + `messages.tags`), which is a separate per channel feature that happens to share the word "tag". To avoid confusion the new vocabulary table is named `upload_tags` and its association `attachment_tags`.

### Changes

**Data model (`src/database.js`, `src/uploadTags.js`)**
- New `upload_tags` table (the vocabulary: display name plus a case folded `name_norm` unique key) and `attachment_tags` table (the association: `message_id`, `rel_path`, `tag_id`), created alongside the existing upload tables, with indexes on `tag_id` and `message_id`.
- New `manage_tags` permission, seeded to the Server Mod role on fresh databases and added to the valid role permission set.
- Tags are deliberately not folded into the FTS index, so `tag:` stays a precise structured filter separate from body text.

**Compose, store, display (Phase 1)**
- A tag bar under the composer attachment previews. Applying an existing tag is open to any uploader; creating a new tag requires `manage_tags`, enforced server side, not just in the UI. New tags commit on message send, so an abandoned draft leaves no orphan vocabulary.
- The tag bar is gated to non DM, non forum channels (DMs are E2E so the server never sees the bytes to index; forum channels have their own topic tag UI).
- Messages carrying tagged attachments render a Tags footer, and each chip is clickable.

**Search (Phase 2, `src/socketHandlers/messages.js`, `app-search.js`)**
- A `tag:` search token with a dropdown picker. The match is a case folded prefix (so `tag:do` finds `dog`, `dogs`, `doghouse`), quoted values support multi word tags, and it is index backed via a subquery on `attachment_tags`. Result rows render the same clickable Tags footer.

**Retroactive editing (Phase 3a, `app-messages.js`, `src/socketHandlers/messages.js`)**
- An "Edit tags" context menu entry on any non DM message that carries an upload. Editing your own message is always allowed (applying existing tags is open, minting a new one still needs `manage_tags`); editing someone else's needs `manage_tags`. `set-message-tags` replaces the full set and broadcasts `message-tags-updated`, so every rendered copy of the footer repaints live.

**Gallery (Phase 3b, `app-ui.js`, `src/socketHandlers/messages.js`)**
- Read only tag chips on each gallery tile, a Tags filter (AND, exact match), and a bulk Append or Replace action in select mode gated to `manage_tags`. Replace with an empty set clears tags behind a second danger confirmation. `get-channel-media` entries now carry their tags, and `bulk-tag-messages` broadcasts `message-tags-updated` per message.

**Admin (Phase 4, `app-admin.js`, `src/socketHandlers/admin.js`, `tags.js`, `uploadTags.js`)**
- Max tags per attachment (1 to 10, default 3) and max tag length (1 to 50, default 20) are `server_settings`, surfaced in the Limits admin section and clamped to the same ceilings on both sides; the apply, edit, and bulk paths read the live limits via `effectiveLimits`.
- A `manage_tags` gated Tags admin panel to add, rename, and delete vocabulary entries. Rename and delete are hard and propagate to every attachment; a rename whose normalized name collides with another tag merges onto the survivor. Both go behind a destructive confirm, and a shared helper rebroadcasts affected footers so open views update live.

**Safety fix for the gallery toolbar (`public/app.html`, `app-ui.js`)**
- Moved the Manage tags control out of the select mode action group (where it sat directly beside Delete selected) to sit next to the Tags filter, so a tag action and the destructive delete are no longer adjacent.
- Scoped Select all to the items visible under the active tag filter (reusing `_filterMediaItemsByTags`) instead of every item in the tab, so it never reaches attachments hidden by the filter.

### Testing
- Normalization, dedupe and cap, the create gate for non `manage_tags` users, case insensitive reuse, and `rel_path` extraction verified against an in memory database.
- Prefix and quoted multi word `tag:` search, per message dedup and grouping, and the retroactive replace and clear paths verified against an in memory database.
- Admin create, pure rename, merge on collision, hard delete, usage counts, length and case insensitive validation verified against the live socket.
- Gallery: tag chips render on tiles, the Tags filter narrows by AND exact match, bulk Append and Replace apply and broadcast, and Replace with no tags routes through the second danger confirm.
- Safety fix: Manage tags renders next to Tags and its Append and Replace dropdown still works, separated from Delete; Select all with a filter active selects only the on screen filtered items, and with no filter behaves as before.
- Full suite matches the clean baseline with zero tagging related regressions.


ancsemi:
Merged, thank you. I read all of it and ran it on a test server first: tagging on send, tag: search with a prefix and with quotes, editing, a rename that merges, and the gates for someone without Manage Tags all behaved. Two small things on top on main: someone who cannot make tags does not see the tag bar until the server has at least one tag, so servers that never use it do not get an empty row, and Reset to Default gives the Server Mod Manage Tags the same as a fresh database does. This covers #5599, which is the request it came from.

PR #5680, opened 2026-09-19
[feat+fix] Frequent tags and retroactive tag bug fix

Bo0sted:
### Summary
Three improvements on top of attachment tagging (#5678):
1. A Frequent tags row under the composer Tags row for one click access to the tags you use most.
2. A guard so a frequent tag that was deleted from the vocabulary cannot be applied silently.
3. A fix for the retroactive Edit tags menu not appearing on a just uploaded message until a refresh.

### Background
The composer only offered tags through the search popup, so re-applying the same handful of tags across many uploads meant typing or searching each time. Separately, the retroactive Edit tags entry was missing on a message you had just posted: it only came back after a page refresh or a channel switch, which looked like the feature was broken.

### Changes
**Frequent tags row (`public/app.html`, `public/js/modules/app-media.js`, `public/css/style.css`, `public/locales/en.json`)**
- New `#tag-frequent-bar` row rendered directly under the Tags row, showing the tags used most on this browser's uploads, most used first.
- The list is stored in `localStorage` (`havenFrequentTags`), not the database, since it is a per person convenience. Each entry keeps a use count and last used time; the row shows the top ten that are not already on the active attachment.
- Counts are recorded when an upload is sent (both the image and general file paths), and the row re-renders whenever the composer tag bar does.
- Clicking a chip applies that tag to the attachment currently selected in the queue, reusing the existing `_applyTagToActive` path.

**Deleted tag guard (`public/js/modules/app-media.js`, `public/locales/en.json`)**
- Because the frequent list lives in storage it can outlive a tag. Before applying a frequent tag the client verifies it still exists via `search-upload-tags`. If it was deleted, the user is shown an error naming the tag, the tag is removed from their storage, and the row refreshes, instead of applying a tag the server would silently drop.

**Retroactive Edit tags menu on fresh uploads (`public/js/modules/app-utilities.js`)**
- `_getMessageAttachments` only read `_lastRenderedMessages`, which holds the last full render and never contains live appended messages, so a just posted upload reported no attachment and the Edit tags entry was gated off until a full re-render.
- It now falls back to the rendered message content in the DOM when the message is not in the last render, scoped to `.message-content` so an author avatar is not mistaken for an upload.

### Testing
- Frequent row appears under Tags once tags have been sent, orders by use count, hides tags already on the active attachment, and hides when the attachment is at its tag cap or the tag bar is not eligible.
- Clicking a frequent chip applies the tag to the selected attachment; the count persists across reloads via localStorage.
- Deleting a tag from the admin panel, then clicking it in the frequent row, shows the error, removes it from storage, and drops it from the row.
- After uploading without tagging, the Edit tags context menu is available immediately on the new message with no refresh or channel switch.


Bo0sted:
Sorry for the PR spam, you happened to merge my original PR right as I was in the process of deploying these changes 😅

Bo0sted:
In regards to [42780ce](https://github.com/ancsemi/Haven/pull/5680/commits/42780cedfc56d3a38aa5cb265574acd84609a0e8)

### Summary
Makes the new frequent tags row honor the tag bar gating added in the attachment tagging follow-up (#5678), so it never lingers after the composer tag bar itself is hidden.

ancsemi:
Merged, and out in 4.10.0. The Edit tags entry showing up on a fresh upload was a good catch.

PR #5681, opened 2026-09-19
Sidebar chrome and the Haven (original) layout switch

Amnibro:
## Summary
- Channels has a + that opens Join, Create and Temporary channel as a sheet, so the two forms stop sitting under the list.
- Direct messages live behind a DMs chip in the footer, with People beside it to open the member panel.
- The home icon opens a menu: Add Server, Manage Servers, Sync, Server settings, App settings.
- Every channel row has Join Voice, which turns into Leave while you are in that call. The header button and the context menu entry are gone.
- Theme, Layout lists Haven (original), Braid and Compact and keeps them exclusive. A file theme can name its layout with `@haven-layout`, and picking Matrix over a Braid theme drops the Braid tokens instead of stacking them.
- Thread mentions count on the threads icon.

Second piece split out of #5658, on current main. The permissions grid is #5675. Ferry gallery, STUN defer, Save Image and the Tauri picture fix are already on main and are not in here; the forum leftovers I still have to check against 4.8 and 4.9.

## Test plan
- [ ] + on Channels opens Join, Create and Temporary as sheets, and each closes after use or with the x
- [ ] DMs chip opens and closes the DM drawer; People chip opens the member panel
- [ ] Home icon menu shows Server settings only to someone with admin settings access
- [ ] Join Voice on a row joins that channel's call; it reads Leave while in it
- [ ] Theme, Layout: Haven (original), Braid and Compact stay exclusive; original brings the header Join Voice back
- [ ] Matrix over a Braid theme keeps the Matrix palette

Full suite passes (369). test/sidebarChrome.test.js covers the chrome, and compactLayout and themeCompatibility cover the layout switch.


ancsemi:
I have read it and had it running on a test server. Since this one changes the default look for everybody (the header buttons, the server rail, where DMs live), I want to sit with it before deciding how it goes in, so it is not in 4.10.0. I will come back to you here.

ancsemi:
I ran this on my own server for a couple of days with the old look one click away, and I am not taking it in as it stands. Some of that is taste and some of it is bugs, so here is what I saw. The + for channels sits inline at the same size as the buttons beside it, which makes creating a channel feel like a minor action when it is the main one there. The dividers between the channel groups are gone, so the list reads as one long column. And there was a large block of dead space under the channels where the DMs used to be, which looks like a layout bug rather than a choice.

Rather than go back and forth on one big PR, I would suggest the route Braid and Compact took: ship this as a layout plugin people can switch on under Theme, Layout, let the people who like it live in it for a while and say what is off, and then we look again at what, if anything, should change for everyone. Any pure fixes hiding in here can come as small PRs on their own. Leaving this open for now so the discussion has a home.

PR #5692, opened 2026-09-23
[fix] Server default theme loads no stylesheet when named without the file: prefix

reverb256:
## What

`default_theme` holds `file:<name>` for a theme that ships as a file. A bare filename is
the same intent — it is what you write by hand when you set the thing. The client tells a
file theme from a built-in name by that prefix, so a bare name is applied as an *unknown
built-in*: the page sets `data-theme="<name>"`, loads no stylesheet, and reports nothing.

The trap is that every server-side surface still looks right. `/api/themes` shows the file
as published, `/api/public-config` returns the default you set, `/themes/<name>.theme.css`
serves 200. The only way to see it is to load the page.

## What this changes

- A name that resolves to a published, compatible theme is served as `file:<name>`,
  whether or not the prefix was written.
- A name that could only be a theme file (`*.theme.css`) and does not resolve to one now
  yields `''`, so a theme you deleted cannot leave an invisible broken default behind.
- Built-in names (`nord`, `haven`, …) and `''` pass through untouched.

`isThemeFilename` already existed and is used one function above for the same reason.

## How it was found

Wiring an Omarchy theme into a Haven server. The setting was written without the prefix;
the login page flashed the right palette in one browser and nothing in another, and every
API I checked said the theme was fine. The behaviour is not documented anywhere a reader
would find it — the prefix requirement lives in the client's branch.

## Verification

`node --test test/themeCompatibility.test.js` — 34 pass, 0 fail, including the new case
(bare filename normalised, unpublished file name rejected, built-ins untouched, `''` and
`null` unchanged) and the existing admin-setting tests, which still pass unchanged.

Only `src/themeMetadata.js`, its test file and one line under `## [Unreleased]` in
`CHANGELOG.md` are touched.


ancsemi:
Merged, thanks! That one would have been a pain to track down from the admin side, since every API said the theme was fine.

